PR #21, opened 2018-11-14
Fix JAVA_HOME setting

lumpfish:
Removes bin directory from end of JAVA_HOME


alexhass:
I think this is not correct... java.exe is in \bin folder and not the folder up. If \bin is removed the java.exe cannot found than and you cannot issue java -version command everywhere in cmd.exe

lumpfish:
Yes, java.exe is in /bin, so that should be added to PATH.
JAVA_HOME is used by applications (admittedly in some cases inconsistently) to locate jdk resources, not necessarily just executables.
For example: https://docs.oracle.com/cd/E19182-01/820-7851/inst_cli_jdk_javahome_t/

alexhass:
I was wrong. I will send a pull asap.

karianna:
Thanks, @lumpfish this helped resolve a number of issues.  Closing this as superseded by #23

Issue #32, opened 2019-01-08
Double-click on JAR doesn't launch the application 


xtbe:
(Moved to here as #684 was closed)

When you double-click on a JAR the application it contains doesn't launch.
This is different behavior compared to the Oracle JDK/JRE. With this version this works.

Tested on Windows 10 x64

todivefor:
This same issue exists for Mac OS X.  With Oracle JDK/JRE, the jar is associated with "jar launcher.app" (15.0.1).  This app will not launch an AdoptOpenJDK created JAR.

PR #34, opened 2019-01-17
Double-click on JAR doesn't launch the application #32

douph1:
Adding registry keys to map .jar extension with javaw.exe -jar

closes: https://github.com/AdoptOpenJDK/openjdk-installer/issues/32


alexhass:
That looks not the recommended way to register an extension via msi. I will share an example later.

gdams:
@alexhass please submit a PR if there is a better way :)

alexhass:
See https://github.com/alexhass/vscode/blob/master/build/wix/code.wxs#L491

PR #43, opened 2019-02-12
Merge cmd script into one/enable cache/check error

douph1:
- Merge multiple cmd script into one
- Add error handling
- Add Cache for faster multilanguage msi build
- Postpone msi validation


gdams:
This PR will require the build scripts in Jenkins to be updated. Please bear this in mind before landing

karianna:
> This PR will require the build scripts in Jenkins to be updated. Please bear this in mind before landing

Where are those build scripts currently?  Do we have them extracted to a GitHub repo?

alexhass:
Caching and reusing MST fIles can cause many troubles. We need to make sure they are recompiled if a translation change and/or the main xml uses a string in a different feature. I‘m interrested how you‘d like to make this a build reliable than...

I‘m also looking forward to the single cmd... that was split as I have no idea how to do without duplicating code. We should think about building for 40 languages...

douph1:
@alexhass

alexhass:
Cool :-) many thanks for this tip.

PR #44, opened 2019-02-12
Fix environment variables

douph1:
- Put jdk/bin at beginning of PATH
- Override JAVA_HOME or JRE_HOME .. dont append to it


alexhass:
Damn copy paste... :-). Changes look good to me.

PR #52, opened 2019-02-20
include JRE in Bundle indentifier

gdams:


pkgbuild/pkgbuild.sh
```diff
@@ -80,16 +80,24 @@ case $JVM in
   openj9)
     IDENTIFIER="net.adoptopenjdk.${MAJOR_VERSION}-openj9.${TYPE}"
     DIRECTORY="adoptopenjdk-${MAJOR_VERSION}-openj9.${TYPE}"
-    /usr/libexec/PlistBuddy -c "Set :CFBundleGetInfoString AdoptOpenJDK (OpenJ9) ${FULL_VERSION}" "${INPUT_DIRECTORY}/Contents/Info.plist"
-    /usr/libexec/PlistBuddy -c "Set :CFBundleName AdoptOpenJDK (OpenJ9) ${MAJOR_VERSION}" "${INPUT_DIRECTORY}/Contents/Info.plist"
+    BUNDLE="AdoptOpenJDK (OpenJ9)"
```
karianna: Is this line needed if the case below actually does the setting?  I notice you don't follow the same pattern for Hotspot...

gdams: It allows for more variants to be added if needed

PR #72, opened 2019-03-14
enhanced comment on config file

douph1:
blank lines are not a problem.


douph1:
@gdams : for your fears about bad editing  :)

PR #75, opened 2019-03-19
Create LICENSE

gdams:


douph1:
This is not the last version from http://www.apache.org/licenses/LICENSE-2.0.txt wich reaplce {} with []

Dont we must apply what explain in appendix ?

   APPENDIX: How to apply the Apache License to your work.

      To apply the Apache License to your work, attach the following
      boilerplate notice, with the fields enclosed by brackets "{}"
      replaced with your own identifying information. (Don't include
      the brackets!)

PR #87, opened 2019-04-17
Add IcedTea-Web as an optional part of the install

charliegracie:
Signed-off-by: Charlie Gracie <charlie.gracie@gmail.com>

### DIFF
--- a/wix/Lang/OpenJDK.Base.en-us.wxl
+++ b/wix/Lang/OpenJDK.Base.en-us.wxl
@@ -18,4 +18,8 @@
 	<String Id="DowngradeDenied">"A later version of [ProductName] is already installed. Setup will now exit."</String>
 	<String Id="FeatureOracleJavaSoftName">JavaSoft (Oracle) registry keys</String>
 	<String Id="FeatureOracleJavaSoftDescription">Overwrites the reg keys HKLM\Software\JavaSoft (Oracle). After uninstallation of AdoptOpenJDK, Oracle Java running from PATH "C:\Program Files (x86)\Common Files\Oracle\Java\javapath" does not work. You need to reinstall it to recreate these registry keys</String>
+	<String Id="FeatureIcedTeaWebTitle">IcedTea-Web</String>
+	<String Id="FeatureIcedTeaWebDescription">Install the IcedTea-Web package</String>
+	<String Id="FeatureJNLPFileRunWithTitle">Associate .jnlp</String>
+	<String Id="FeatureJNLPFileRunWithDescription">Associate .jnlp files to run with IcedTea-Web</String>
 </WixLocalization>
--- a/wix/Build.OpenJDK_generic.cmd
+++ b/wix/Build.OpenJDK_generic.cmd
@@ -68,6 +68,7 @@ REM
 REM Cultures: https://msdn.microsoft.com/de-de/library/ee825488(v=cs.20).aspx
 SET PRODUCT_SKU=OpenJDK
 SET PRODUCT_VERSION=%PRODUCT_MAJOR_VERSION%.%PRODUCT_MINOR_VERSION%.%PRODUCT_MAINTENANCE_VERSION%.%PRODUCT_PATCH_VERSION%
+SET ICEDTEAWEB_DIR=.\SourceDir\icedtea-web-image
 
 
 REM Generate platform specific builds (x86-32,x64)
@@ -175,7 +176,14 @@ FOR %%A IN (%ARCH%) DO (
 	    GOTO FAILED
 	)
 
-    REM Build without extra Source Code feature
+	REM Build without extra Source Code feature
+
+	ECHO HEAT
+	"!WIX!bin\heat.exe" dir "!ICEDTEAWEB_DIR!" -out Files-IcedTeaWeb.wxs -t "!SETUP_RESOURCES_DIR!\heat.icedteaweb.xslt" -gg -sfrag -scom -sreg -srd -ke -cg "IcedTeaWebFiles" -var var.IcedTeaWebDir -dr INSTALLDIR -platform !PLATFORM!
+	IF ERRORLEVEL 1 (
+		ECHO "Failed to generating Windows Installer XML Source files for IcedTea-Web (.wxs)"
+		GOTO FAILED
+	)
 
 	ECHO HEAT
 	@ECHO ON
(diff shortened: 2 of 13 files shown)


charliegracie:
This PR is adding support to make IcedTea-Web be an optional part of the AdoptOpenJDK installer. Currently the changes expect and IcedTea-Web artifact to be present and the installer build will fail if it does not exist. 

Is this the correct behaviour? Or should the build just create the installer without optionally adding IcedTea-Web if it does not exist?

charliegracie:
I have tested this with the latest OpenJ9 JDK 8 release. I will test it with HotSpot JDK 8 shortly.

douph1:
@charliegracie , please build a installer with job
"standalone_create_installer_windows" (pointing to  refs/pull/87/head ) to show us the result when you think it is ready

>the installer build will fail if it does not exist.
>Is this the correct behaviour?

Yes, We start to include icedtea into installer so it must always be included

charliegracie:
I do not see that job on Jenkins. Maybe I do not have rights to launch it.

karianna:
> I do not see that job on Jenkins. Maybe I do not have the rights to launch it.

I don't see the job there either - https://ci.adoptopenjdk.net/view/IcedTea-Web/ contains the IcedTea-Web builds that we know of.  Perhaps @douph1 was suggesting to create a new job?

douph1:
> > I do not see that job on Jenkins. Maybe I do not have the rights to launch it.
> 
> I don't see the job there either - https://ci.adoptopenjdk.net/view/IcedTea-Web/ contains the IcedTea-Web builds that we know of. Perhaps @douph1 was suggesting to create a new job?

I speeking about the Installer job 
I have let a msg this morning on slack to @charliegracie  wich seem absent.

I have run the standalone job
>heat.exe : error HEAT5052 : The directory 'C:\Users\jenkins\workspace\build->scripts\release\standalone_create_installer_windows\wix\SourceDir\icedtea-web-image' could not be >found.
>"Failed to generating Windows Installer XML Source files for IcedTea-Web (.wxs)"
>Build step 'Execute Windows batch command' marked build as failure

I dont know exactly how it works but I think we must configure installer job to get artifact (icedtea-web/icedtea-web-*.win.bin.zip ? ) from icedtea-web_build_x86-64_windows

so we can re-use installer job with artifact created by icedtea job

charliegracie:
Sorry I am unavailable on Slack until later today.  Yes however the installer job gets an OpenJDK zip file it should get the IcedTea-Web zip file from the Jenkins job.  Once you put the zip in the SourceDir with the OpenJDK zip everything else should work the same as before.

douph1:
https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/test_installer_windows/1/
Finished: SUCCESS

and I have tested OpenJDK8-jre_x64_windows_hotspot-8.0.202.08.msi for installer work as expected and JNLP are well associated with javaws.

karianna:
I'm OK with us adding this in but only once we have a stable 1.8.1 ITW released.

wix/Lang/OpenJDK.Base.en-us.wxl
```diff
@@ -18,4 +18,8 @@
 	<String Id="DowngradeDenied">"A later version of [ProductName] is already installed. Setup will now exit."</String>
 	<String Id="FeatureOracleJavaSoftName">JavaSoft (Oracle) registry keys</String>
 	<String Id="FeatureOracleJavaSoftDescription">Overwrites the reg keys HKLM\Software\JavaSoft (Oracle). After uninstallation of AdoptOpenJDK, Oracle Java running from PATH "C:\Program Files (x86)\Common Files\Oracle\Java\javapath" does not work. You need to reinstall it to recreate these registry keys</String>
+	<String Id="FeatureIcedTeaWebTitle">IcedTea-Web</String>
+	<String Id="FeatureIcedTeaWebDescription">Install the IcedTea-Web package</String>
```
tellison: While we know this feature by the internal project codename 'IcedTea-Web",  I think it may be better to refer to the install option in the UI by the technology name "JNLP".  WDYT?

SueChaplain: Note intent to move the user doc for installers to the website here: https://github.com/AdoptOpenJDK/openjdk-website/pull/479 Still some work to do if we are using a new INSTALLLEVEL and changing the name :-/

SueChaplain: What is the status here please? 

- Are we changing the name from IcedTea-Web to JNLP?
- Are we creating a new INSTALLLEVEL=3 for IcedTea-Web or will the following features be part of INSTALLLEVEL=2: FeatureIcedTeaWeb,FeatureJNLPFileRunWith

@tellison @charliegracie @karianna @douph1 ?? Anyone else?

karianna: I'm not close enough to the technology from an end user perspective.  I'd argue that we should follow the experience that the user has today with the existing Oracle Java / Java Webstart and/or IcedTeaWeb installation.  Do we have any end users who can check those settings?

kelderek: End user here.  The Oracle Java installer from java.com has java webstart functionality built in and default, which I use for connecting to my Dell DRAC out of band server management.  No separate installs or special options to choose in the Oracle installer.  jnlp files are automatically associated.  I suspect java webstart is rarely used outside of sysadmin circles these days, so I think a separate INSTALLLEVEL is fine.  If you want it to match the Oracle installer, though, IcedTea-Web and .jnlp file association should be default.

douph1: > * As an optional install the new feature should be available in a new INSTALLLEVEL=3

This would mean that selecting INSTALLLEVEL=3 would install all other optionnal features JAVA_HOME/JavaSoft key/PATH.. and maybe we dont want all juste because we want ITW.

I think INSTALLLEVEL=1 is for JDK/JRE ( required )
and all other are LEVEL=2
As ITW don't depend on JAVA_HOME or PATH or  JavaSoft they must not be greater than 2.

INSTALLLEVEL=1 is minimal
INSTALLLEVEL=2 is all optionnal ( Full install )

If you need finer grained install use ADDLOCAL with list of features

douph1: >I suspect java webstart is rarely used outside of sysadmin circles these days

I dont kniw what "rarely" is but
my company ( and probably many others ) are selling sofware available by JavaWebStart to many end-user ( more than 8000 uniq user this last month ) and this is not sysadmin software

kelderek: I am happy with it just being an ADDLOCAL option and not a new INSTALLLEVEL.  Really I just want unattended install capability, which an ADDLOCAL option for ITW will give me just like the current version does with setting the JAVA_HOME variable. 
No offense was meant with the "rarely used" comment, I was just speaking to my personal experience.  I thought it was a deprecated technology, I didn't know it was still in active use.

chrisbrookes: Our company uses it for sysadmin **and** we sell software to customers that uses it. As far as Oracle is concerned, it's deprecated, but I suspect quite a lot of enterprises that use/sell Java Desktop software use it. Ideally I would prefer it installed by default, but not a huge problem if we have to document a "tick this box" - I understand that not everybody necessarily wants it installed.

SueChaplain: It would be good to get some closure here. My vote is with @douph1:

>INSTALLLEVEL=1 is minimal
>INSTALLLEVEL=2 is all optionnal ( Full install )

>If you need finer grained install use ADDLOCAL with list of features


Who else is voting? ;-)
Would there be new changes needed to implement this?

kelderek:
I definitely _don't_ want HKCU - due to regulations imposed on us, installing things per-user is a major pain and we want deployed software to apply to all users by default.

douph1:
> I definitely _don't_ want HKCU - due to regulations imposed on us, installing things per-user is a major pain and we want deployed software to apply to all users by default.

Dont worry, this is not the default behaviour.
We use ALLUSERS=2
See https://docs.microsoft.com/en-us/windows/desktop/msi/allusers

douph1:
PS: @kelderek I dont have mentionned HKCU but HKMU
http://wixtoolset.org/documentation/manual/v3/xsd/wix/simple_type_registryroottype.html
https://docs.microsoft.com/fr-fr/windows/desktop/Msi/registry-table (- 0x001)

kelderek:
> 
> 
> PS: @kelderek I dont have mentionned HKCU but HKMU
> http://wixtoolset.org/documentation/manual/v3/xsd/wix/simple_type_registryroottype.html
> https://docs.microsoft.com/fr-fr/windows/desktop/Msi/registry-table (- 0x001)

Ok, thanks.  I have never seen HKMU before so I assumed it was a typo.  I did Google HKMU before posting, but the first page had no relevant results.  Thanks again!

charliegracie:
@douph1 I have rebased on master and updated the registry keys to `HKMU`.

douph1:
I have trigger a new build with last change:
https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/test_installer_windows/2/

douph1:
@charliegracie  what do you think of letting FeatureIcedTeaWeb at INSTALLLEVEL=2
but moving FeatureJNLPFileRunWith at INSTALLLEVEL=3 because it one level deeper in the tree

And doesn't we spacing the INSTALLLEVEL to allow adding feature in the midle ?

INSTALLLEVEL=1
INSTALLLEVEL=10
INSTALLLEVEL=20

So later we can add Update-notifier at LEVEL 5 or 15 ( between existing )

charliegracie:
I think changing the JNLP association to LEVEL=3 makes sense. I will change it now.  

I think changing all of the LEVELs like you mentioned may make sense but probably best left for a future PR.

PR #90, opened 2019-04-21
 Add tasks to upload packages to Bintray

aahlenst:
This pull requests adds the necessary tasks to upload the AdoptOpenJDK Debian and RPM packages for all supported distributions and architectures to the AdoptOpenJDK Bintray repositories. The whole process including package creation is automated and can be defined in code. Automatic publication is opt-in (i.e. it can be disabled for a manual review of releases). It's also possible to publish nightlies and releases to different repositories. Everything's Gradle and Java (some tests are in Groovy in order to preserve my sanity) so it should run on all build platforms.


linux/buildSrc/build.gradle
```diff
@@ -1,7 +1,80 @@
+// Plug-in order matters. The groovy plug-in must appear after java-gradle-plugin or some tests
+// won't be picked up.
+plugins {
```
karianna: Versions of the plugins?  Could do with Ben Manes deps checker as well.

linux/buildSrc/checkstyle.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
```
karianna: Not added to the build.gradle?

aahlenst: It is, see `linux/buildSrc/build.gradle` (L22-L27). It's only there for the plug-in sources. Or you do mean something else?

linux/buildSrc/build.gradle
aahlenst: Those are bundled plug-ins (https://docs.gradle.org/current/userguide/plugin_reference.html) and as far I understand automatically versioned together with Gradle.

karianna: Yep, we should still probably try to version the plugins independent of the Gradle version (repeatable builds).

linux/buildSrc/checkstyle.xml
karianna: I missed it, my bad!

linux/buildSrc/build.gradle
aahlenst: I'm totally for repeatable builds. As far as I understand it, it's already as repeatable as it gets. The Gradle version is fixed via Gradle Wrapper (currently 5.4). As a result the bundled plugins are at whatever version Gradlew 5.4 shipped with. Or am I mistaken here?

If I specify a version there, we have to remember to update those versions if we change the Gradle version. If someone uses a different Gradle version, that's unsupported in any case.

karianna: yeah fair enough

Issue #93, opened 2019-04-24
Double-click on JAR doesn't launch the application - Mac OS X


todivefor:
Issues 32 and 34 were merged and closed.  Those issues were for Windows.  The same issue is true for MAC OS X.  For Oracle JDK created JAR files, they are launched with "jar launcher.app" (15.0.1) found at /System/Library/CoreServices/Jar Launcher.app.  Minor inconvenience, but a difference.  I have found this with all AdoptOpenJDK releases on Mac OS X 10.14 and 10.13.

gdams:
@todivefor any idea how this would be associated?

gdams:
So I've created a .jar file on my mac and when I double click it, it opens with jar launcher app. That seems to be working for me.

gdams:
Or is this to do with making jar launcher.app use adoptopenjdk Java rather than another vendor?

gdams:
running a test build with this fix: https://github.com/AdoptOpenJDK/openjdk-installer/pull/95

gdams:
@todivefor could you please confirm the fix for me by running the following commands making sure that the **INPUT_DIRECTORY** variable is set correctly:

```bash
INPUT_DIRECTORY=/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk
/usr/libexec/PlistBuddy -c "Add :JavaVM:JVMCapabilities:1 string JNI" "${INPUT_DIRECTORY}/Contents/Info.plist"
/usr/libexec/PlistBuddy -c "Add :JavaVM:JVMCapabilities:2 string BundledApp" "${INPUT_DIRECTORY}/Contents/Info.plist"
```

todivefor:
@gdams that did not work.  I first tried it just like you said and it seemed to work.  I then tried it on INPUT_DIRECTORY=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk and it did not work there.  I went back to adoptopenjdk-8.jdk and removed the above fix and re-tryed.  It worked, so my statement all AdoptOpenJDK releases was false.  Double click on 1.8.0_192-b12 works, but 11.0.3+7 does not.

gdams:
hmmm okay, do you know if an Oracle Java 11 works with this?

todivefor:
I do not, but I can give it a try.

todivefor:
[...]

PR #95, opened 2019-04-25
associate Java binaries with jar macOS

gdams:


karianna:
I'll merge once we've proven #93 is resolved by this.

PR #96, opened 2019-04-29
Only depend on libfreetype6 instead of libfreetype6-dev.

nudgegoonies:
libfreetype6-dev is a development dependency which would only be used, if AdoptOpenJDK compiles something to link to it either during or after package installation. But this does not seem to be the case.

### DIFF
--- a/linux/deb/config/jdk-10-hotspot/dependencies.txt
+++ b/linux/deb/config/jdk-10-hotspot/dependencies.txt
@@ -2,7 +2,7 @@ ca-certificates-java
 java-common
 libasound2
 libc6
-libfreetype6-dev
+libfreetype6
 libx11-6
 libxext6
 libxi6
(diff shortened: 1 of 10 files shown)


aahlenst:
We might be able to drop the freetype dependency completely because a `libfreetype.so` is shipped in the `lib` folder, both in the JDK and JRE. Do you have the time/capacity to test this?

nudgegoonies:
I tried with 8 and 11 and both versions load the included freetype library. So i think the dependency can be removed safely:
```
nudge@nudge-laptop:~$  $ sudo lsof -p 13712 | grep freetype
java    13712 nudge  mem    REG              253,1  3277887   2884104 /usr/lib/jvm/adoptopenjdk-8-hotspot-amd64/jre/lib/amd64/libfreetype.so.6
nudge@nudge-laptop:~$  $ sudo lsof -p 13996 | grep freetype
java    13996 nudge  mem       REG              253,1    783560 2902045 /usr/lib/jvm/adoptopenjdk-11-hotspot-amd64/lib/libfreetype.so
```

aahlenst:
Great, thanks a lot!

nudgegoonies:
Thank you!

PR #108, opened 2019-05-10
Revert "enable capability to install per user + doc"

alexhass:
Reverts AdoptOpenJDK/openjdk-installer#107

### DIFF
--- a/wix/Main.hotspot.wxs
+++ b/wix/Main.hotspot.wxs
@@ -9,7 +9,7 @@
     <Property Id="ARPHELPLINK" Value="!(loc.ProductHelpLink)" />
     <Property Id="ARPURLINFOABOUT" Value="!(loc.ProductUrlInfoAbout)" />
     <Property Id="ARPURLUPDATEINFO" Value="!(loc.ProductUrlUpdateInfo)" />
-    <Property Id="ALLUSERS" Value="2" /> <!-- Allow per User or per Machine setup -->
+    <Property Id="ALLUSERS" Value="1" />
     <Property Id="ARPPRODUCTICON" Value="logo.ico" />
     <Property Id="ORACLE_VERSION_KEY" Value="$(var.OracleVersionKey)" />
     <Property Id="ORACLE_JAVASOFT_BASE_KEY" Value="$(var.OracleJavasoftBaseKey)" />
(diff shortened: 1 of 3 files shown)


karianna:
OK let's have a discussion around this.  I assume there's a reason why you prefer a system-wide install vs a user install?  I'm personally used to letting each logged in user choose what version of the software they wish to use.

karianna:
Also, there's a merge conflict.

douph1:
@alexhass was speeking about https://www.hass.de/content/developers-best-practice-windows-user-profiles this in mind and wixtoolset/issues#5908

I don't see the point, the link is just related to path where to install when you are on per user install

>If a software does not work this way and/or if permanent data is saved in LocalAppData aka AppData\Local the Application has a bug that need to be fixed.

We don't force user to install into LocalAppData and so it is not a bug.
Maybe @alexhass you have seen this is the default folder proposed by wix/msi installer is LocalAppData
and if it is a real problem as you suggest we can proably set another default folder for per-user install.
Any proposal @alexhass  ?

for https://github.com/wixtoolset/issues/issues/5908
I dont see the point either .. ProgramFilesFolder is for machine ( or admin ) only on x86 and 64 .. this is not ever linked to per-user install

douph1:
@karianna 
> I'm personally used to letting each logged in user choose what version of the software they wish to use.

Yes it can be a better experience for user to let him choose with UI instead of just by cmdline or msi transform.

If you use a system or admin account there is probably the need to allow him to choose.
If you are not priviliged account, there is no choise available, you can only installer per-user or not.

The only problem seen and documented at this time is that per-user can't write HKLM and so can't write JavaSoft reg key and it is documented.
 ( and maybe the default dest folder proposed by wix/msi in per-user install , I will check )

douph1:
https://sourceforge.net/p/wix/feature-requests/656/

>But I agree it makes sense to support both local and roaming profiles

alexhass:
As admin I can only never allow a user to install software per user. 

Typical software allows me to set a GPO to prevent per user installs e.g. all Google Software products. The only place software can be installed per user is local app data. This folder and subfolders will be killed and the application will be broken and half removed (files gone) rest stays as on every user logoff. No idea why you implemented per user installer. I would never create one.

Automatisms are always bad, too. The user cannot choose a folder and is not notified to switch to an admin user like normally. Normally a user gets an UAC, enters credentials and than software is installed per machine.

If he has no permission he is not allowes to install software and need to go to an admin.

Per user installs are now causing serious security issues as there is an unmaintained and forgotten java install on an enterprise system installed. This will cause serious security issues to companies (and private people).

Never install software per user, please!

gdams:
@alexhass surely in enterprise systems where this matters you can just block users from installing any form of MSI?

alexhass:
No, that I’m aware that this is possible. Some applications like Microsoft Office may try to complete an installation. Not sure why, but it happens. It is also not the point. Per user installs are always bad.

douph1:
>No idea why you implemented per user installer

This is not per-user installer.. this is per machine **or** per-user and default always per machine.

and because it was asked
https://github.com/AdoptOpenJDK/openjdk-installer/issues/103

and documented by Microsoft :
first § https://docs.microsoft.com/en-us/windows/desktop/msi/allusers
https://docs.microsoft.com/en-us/windows/desktop/msi/installation-context
https://docs.microsoft.com/en-us/windows/desktop/msi/single-package-authoring

and wix supported too HKMU etc

>Automatisms are always bad

This is not automatism, as said default behaviour is always the same.
You must explicitly set an property to use per user install, this is not automatism

> The user cannot choose a folder

He can ..
Have you tried ?

>and is not notified to switch to an admin user like normally

Have you tried ? by default this is always per-machine and user is notified

>Normally a user gets an UAC, enters credentials and than software is installed per machine.

It is always the case by default if user dont use new property

>Per user installs are now causing serious security issues as there is an unmaintained and forgotten java install on an enterprise system installed

Not more or less than a uncompressed zip file from AdoptOpenJdk wich is unmaintained.
You can't prevent security issue in your environnement by asking all over the world to not build per-user installer.

So block install for all MSI and whitelist only what needed/allowed by "publisher" exemple  "Microsoft corp" for the exemple you gabe (Microsoft Office)

There is many portable bin files today, and yes, if allowed to be run it is a security issue.
You can also deny exec from binaries inside LocalAppData


Maybe if we finally decided to keep this feature we can put notice on the doc or installer itself :
Install per user is discouraged and not supported .. use it as your on risk .. no support will be done on it.

If we build/publish two installer ? One per machine and one per user may sweet you ?

@alexhass you did not convince me, but be sure I did not want to do bad/wrong things by doing this. I just wanted to answer a feature request, it was not too complicated, so I did it quickly.

My fault was probably to not wait longer and more feedback before merging

alexhass:
> This is not per-user installer.. this is per machine or per-user and default always per machine.

This is half true. If the users is detected as non admin it starts installing per user. But the user could have Admin password. But you do not tell him and install silently per user. You may also install the same version that is already per machine installed - per user just because the user has no admin privilege. Users have no clue what they are doing. If someone write a documentation - please install blah blah version they are doing this - also if outdated. Now my machine installer runs via AD and installs a newer version. The older per user installation stays as is is never updated. If I uninstall Java because it is dead on desktops the machine install is gone, but the security problem (outdated version) stays on the user and the user is attackable by you know - everything.

Software installation should be done by Admins only. People who know what they are doing. With an auto per user installer you undermine security and software deployment rules of companies. You take away the control and in parallel you do not provide a way to block per user installations.

My users do not have space in `%USERPROFILE%\Application Data\Microsoft\Installer\{ProductCode GUID}`. AppData is limited to 4GB in `%APPDATA%`. If everything installed per user they have 30GB of installers roaming. Have you kept in mind that APPDATA is a redirected folder? I'm currently not sure if `%USERPROFILE%\Application Data` is `local app data` or the `roaming app data` folder or the` redirected app data` folder.

> You must explicitly set an property to use per user install, this is not automatism

Does it mean the per user installer is not trying to detect if the user has permission if someone runs the setup? This is how I'm reading the current code. It tries to detect if admin or user and if no admin it runs the per user setup. Do you mean I need to provide ALLUSERS=2 on command line and otherwise the per user installer is **never** used?

> The user cannot choose a folder
> 
> He can ..
> Have you tried ?

No, there seems no final build yet with the feature.

> One per machine and one per user may sweet you ?

No, I do not see a reason to provide per user installers at all. This is only bad design and most developers forgotten about roaming profiles and the delete feature. That is why I linked the article of mine. Every logoff causes typically application loss. I do everything to block per user installers. Here I cannot. Can you check an additional registry key, so that we admins can add a GPO to block per user installs like Google does for all their products, please?

I'm not blocking MSI setups... I would love to block NSIS and InnoSetup, but this is also not possible in general - just with software restrictions, but nothing is really reliable. 

I'm glad that per user installers are and edge cases as of now. vscode is the most worst project around. You only make things more worse for enterprise deployment.

karianna:
OK, so it sounds like we need to provide an option/registry key or something to allow folks like @alexhass to install in the manner that he requires.

I think there is a cultural difference between 'home' or small business users who may have a few user accounts on a machine and want to install per user vs perhaps large/strict corporate environments where per-machine installations are preferred.

douph1:
Sorry, I repeat , one more time
>If the users is detected as non admin it starts installing per user

No, not on my test desktop and not on my terminal server

It doesnt work as you think it does ( see why below )

>But you do not tell him and install silently per user

Again , no, If no property is given there is no per user install

...

> you do not provide a way to block per user installations

right but Microsoft does with policies as my corp does it



>Does it mean the per user installer is not trying to detect if the user has permission if someone runs the setup?

Yes



> It tries to detect if admin or user and if no admin it runs the per user setup. Do you mean I need to provide ALLUSERS=2 on command line and otherwise the per user installer is never used?


No you must use MSIINSTALLPERUSER=1 or ALLUSERS=""  to activate per-user installer (ALLUSERS=2 is set in Wix xml with `<Property Id="ALLUSERS" Value="2" />` but MSIINSTALLPERUSER is not set within wix xml )

and otherwise the per user installer is never used.


https://docs.microsoft.com/en-us/windows/desktop/msi/single-package-authoring

Because I have not set  ALLUSERS=2 **AND** MSIINSTALLPERUSER=1

the dual-purpose package is not per-user by default.

If I did that, then yes, the package would have been a silent per-user installation by default.
By not setting MSIINSTALLPERUSER property in Wix XML I have not change the default behaviour

And yes in this sense I don't have followed the "Development Guidelines" linked below to not change behaviour.

Some things about security before adding some new reg key.
I have try to install per user (as a non-privilegied user) on our managed(by another team) terminal server which enforce security rules.
I'm logged in as a standard terminal server user and I can try to run msi per user but it is denied before the end when I click last button "install" with a message :
translate from french to english : "The system administrator has set policies to prevent this installation."

**So prevent per user install is already a feature powerd by Microsoft without adding a special reg key**

Yes, I agree, this is not the point, but Microsoft give some policies to manage it
https://docs.microsoft.com/en-us/windows/desktop/Msi/machine-policies
    https://docs.microsoft.com/en-us/windows/desktop/Msi/disablemsi
    https://docs.microsoft.com/en-us/windows/desktop/Msi/disableuserinstalls

PS: the default folder suggested during UI was :
C:\Users\<login>\AppData\Local\Programs\AdoptOpenJDK\jre-8.0.212.03-hotspot\

During UI setup I can easly change it to a folder with more space where I have access (my network drive mapped as a letter )

but the domain policy always reject installation.
"The system administrator has set policies to prevent this installation."


>Can you check an additional registry key

Why not, I dont see any drawback to add just a reading a key and conditional step , feel free to make a PR

alexhass:
> PS: the default folder suggested during UI was :
> C:\Users<login>\AppData\Local\Programs\AdoptOpenJDK\jre-8.0.212.03-hotspot\

Ok. We should all be aware that this is getting lost on logoff if roaming profiles are enabled.

> No you must use MSIINSTALLPERUSER=1 or ALLUSERS="" to activate per-user installer (ALLUSERS=2 is set in Wix xml with <Property Id="ALLUSERS" Value="2" /> but MSIINSTALLPERUSER is not set within wix xml ) and otherwise the per user installer is never used.

Than I'm fine. Sorry for the pointless discussion.

I never heard of https://docs.microsoft.com/en-us/windows/desktop/Msi/disableuserinstalls. Many thanks for sharing this. I will enabled it asap. With this solution in place I'm fine if it works as it already does. Adding another registry key makes no sense to me than, too. That is waste of time.

karianna:
OK, I think we've come to a conclusion, thank you both for digging into this!

@alexhass Are you OK if I close this PR?
@douph1 Do we need to raise a separate PR or are we OK now?

alexhass:
Let‘s close it.

douph1:
I'm OK. I close.
If someone think that we must extend the readme/website to encourage user to use per machine setup, feel free to create a new PR

PR #110, opened 2019-05-11
Update supported Linux distributions

aahlenst:
* Fedora: Drop 28, add 30
* CentOS, RHEL: Add 8
* Ubuntu: Drop Trusty


rodionmoiseev:
@karianna 
Hi! I was wondering if this update implies that RHEL 8 and CentOS 8 can also appear in the supported_platforms.html page?
Thanks!

karianna:
> @karianna
> Hi! I was wondering if this update implies that RHEL 8 and CentOS 8 can also appear in the supported_platforms.html page?
> Thanks!

It would do yes, can you raise a website issue for us?

rodionmoiseev:
@karianna Thanks for the prompt reply!

Sure, will raise the issue. 

Thanks

Issue #105, opened 2019-05-06
Debian packages cannot be installed together with ca-certificates-java


nudgegoonies:
Because of the dependency of adoptopenjdk-* to ca-certificates-java in debian they are installed at the same time. For some reason both cannot be installed at the same time because ca-certificates-java needs a ready installed java to execute java but ca-certificates-java seems to be installed first so there is no java executable yet.

We have to look how openjdk-8-jre-headless does the magic in the debian package. It seems that the problem is this:
```
setup_path()
{
    for jvm in java-7-openjdk-$arch java-7-openjdk \
               oracle-java7-jre-$arch oracle-java7-server-jre-$arch oracle-java7-jdk-$arch \
               java-8-openjdk-$arch java-8-openjdk \
               oracle-java8-jre-$arch oracle-java8-server-jre-$arch oracle-java8-jdk-$arch \
               java-9-openjdk-$arch java-9-openjdk \
               oracle-java9-jre-$arch oracle-java9-server-jre-$arch oracle-java9-jdk-$arch \
               java-10-openjdk-$arch java-10-openjdk \
               oracle-java10-jre-$arch oracle-java10-server-jre-$arch oracle-java10-jdk-$arch \
               java-11-openjdk-$arch java-11-openjdk \
               oracle-java11-jre-$arch oracle-java11-server-jre-$arch oracle-java11-jdk-$arch \
               java-12-openjdk-$arch java-12-openjdk \
               oracle-java12-jre-$arch oracle-java12-server-jre-$arch oracle-java12-jdk-$arch \
               java-13-openjdk-$arch java-13-openjdk \
               oracle-java13-jre-$arch oracle-java13-server-jre-$arch oracle-java13-jdk-$arch \
               java-14-openjdk-$arch java-14-openjdk \
               oracle-java14-jre-$arch oracle-java14-server-jre-$arch oracle-java14-jdk-$arch \
               java-15-openjdk-$arch java-15-openjdk \
               oracle-java15-jre-$arch oracle-java15-server-jre-$arch oracle-java15-jdk-$arch \
               java-16-openjdk-$arch java-16-openjdk \
               oracle-java16-jre-$arch oracle-java16-server-jre-$arch oracle-java16-jdk-$arch \
               java-17-openjdk-$arch java-17-openjdk \
               oracle-java17-jre-$arch oracle-java17-server-jre-$arch oracle-java17-jdk-$arch; do
[...]

PR #111, opened 2019-05-12
Remove ca-certificates-java dependency from Debian packages

aahlenst:
Its circular dependency with the JDK packages makes the AdoptOpenJDK packages uninstallable on Debian (see #105). Trying to somehow get it to install anyway would result in brittle packages because apt/dpkg provides no guarantees how dependency cycles are going to be resolved.

Without ca-certificates-java, the JDK tools will use the default truststore shipped with the JDK.

This is most likely just an interim fix for #105. I plan to bring back the functionality of `ca-certificates-java` without having to install `ca-certificates-java`.

### DIFF
--- a/linux/deb/config/jdk-10-hotspot/dependencies.txt
+++ b/linux/deb/config/jdk-10-hotspot/dependencies.txt
@@ -1,4 +1,4 @@
-ca-certificates-java
+ca-certificates
 java-common
 libasound2
 libc6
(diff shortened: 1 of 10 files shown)


karianna:
@aahlenst Perhaps check with @johnoliver on Slack and make sure that it doesn't clash with the backported CA certs that we bundle.

PR #121, opened 2019-05-23
Only add icedtea-web to jdk8 msi installer

lumpfish:


lumpfish:
@douph1 , @gdams - is there some way I can test his change using the AdoptOpenJDK build machines with the Wix setup?

wix/Build.OpenJDK_generic.cmd
```diff
@@ -178,7 +178,7 @@ FOR %%A IN (%ARCH%) DO (
 
 	REM Build without extra Source Code feature
 
-    IF !PLATFORM! == x64 (
+    IF !PLATFORM! == x64 IF !PRODUCT_MAJOR_VERSION! == 8 (
```
karianna: I would nest that 2nd IF to match the style of the rest of this file

lumpfish: Done

lumpfish:
I've changed the compound IF to a nested IF as per @karianna 's comment.

PR #125, opened 2019-06-05
Correct debian provides to include headless and non headless variants…

nudgegoonies:
… because there are no adoptopenjdk headless builds.

OpenJDK has non-headless and headless packages. They deduplicate as much as possible and so their jdk and jre packages depend on their jdk-headless and jre-headles counterparts and the jdk and jre packages only add the graphic libraries and add Xorg dependencies for them. AdoptOpenJDK does not build headless packages. Because the packages always include the graphic libraries and dependencies they mus not only provide the headless packages but also the non-headless packages.

 We had bug with a package that required java8-runtime that could not be installed with adoptopenjdk.


nudgegoonies:
This is the related issue: https://github.com/AdoptOpenJDK/openjdk-installer/issues/126

nudgegoonies:
Thank you for the finding. I should have looked for the amount of files. 5 Java versions with 4 files each equals 20 files. I added both to my commit.

PR #127, opened 2019-06-08
Add tasks to remove outdated nightly builds

aahlenst:
The preconfigured tasks remove all artifacts that are older than 7 days.

Invoke with `./gradlew cleanupNightlyBuilds`. It requires the usual properties:

```
ARTIFACTORY_USER=
ARTIFACTORY_PASSWORD=
ARTIFACTORY_REPOSITORY_DEB=deb-nightly
ARTIFACTORY_REPOSITORY_RPM=rpm-nightly
```

### DIFF
--- a/linux/build.gradle
+++ b/linux/build.gradle
@@ -57,6 +57,11 @@ tasks.register("upload") {
     description = "Uploads all linux packages"
 }
 
+tasks.register("cleanupNightlyBuilds") {
+    group = "upload"
+    description = "Removes all outdated nightly builds"
+}
+
 def getJdkDistributionType() {
     return hasProperty("JDK_DISTRIBUTION_TYPE") ?
         DistributionType.valueOf(JDK_DISTRIBUTION_TYPE.toUpperCase(Locale.US)) :
--- a/linux/buildSrc/src/functTest/groovy/net/adoptopenjdk/installer/RemoveObsoletePackagesTest.groovy
+++ b/linux/buildSrc/src/functTest/groovy/net/adoptopenjdk/installer/RemoveObsoletePackagesTest.groovy
@@ -1,0 +1,205 @@
+package net.adoptopenjdk.installer
+
+import okhttp3.Credentials
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.RecordedRequest
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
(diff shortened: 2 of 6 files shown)


karianna:
Why do we need to do this?

aahlenst:
In the last few weeks, more than 200 GB of nightly builds piled up in Artifactory. It‘s unlikely that anybody needs those. But they occupy storage and ultimately cost money.

karianna:
@aahlenst OK, but to be clear in the Web UI and API that means those installers will no longer be available (> 7 days old) correct?  We'd need to account for that.

aahlenst:
As far as I know no Linux nightly builds are available through the AdoptOpenJDK website or the API. https://adoptopenjdk.net/nightly.html?variant=openjdk11&jvmVariant=hotspot doesn't list them and `curl -L 'https://api.adoptopenjdk.net/v2/info/nightly/openjdk11?openjdk_impl=hotspot' | grep deb` yields no results.

If we want to make the RPM and DEB packages available on the website and the API, I think we should put them where all the other packages are stored (i.e. GitHub). Bintray/Artifactory are a convenient distribution mechanism for package managers because they make all the metadata generation and most of the signing for us.

karianna:
@gdams I think we should hook the API into picking these up, but it does raise the larger question of how many we should hold.  Might be a TSC issue actually.

gdams:
As this only applies to Nightly builds I'm happy in principle to cleanup builds over 7 days old.

karianna:
@gdams Should we then not apply that policy to Windows and Mac Installers to be consistent?

gdams:
storage is not an issue in github releases so I think not.

karianna:
OK but this will create an inconsistency for our end users when we make the Linux installers available through our API and website.  We'll need to document this decision clearly somewhere.

@aahlenst Can you add some docs to the README.md for this repo - at least that way we'll capture the decision here.

aahlenst:
In order to document it properly: Will the releases and nightly builds be exposed through the API/Website? And if so, where will they be stored?

karianna:
> In order to document it properly: Will the releases and nightly builds be exposed through the API/Website? And if so, where will they be stored?

They will be in due course yes - @gdams Do we have open issues for this in API and Website?

gdams:
@karianna I'm not sure that they will be? I don't see much value in adding the releases to the API.

karianna:
Why not - don't we want the Linux installers available via API and website, else users won't discover them right?  Or is this a case of us promoting Artifactory as an APT and RPM repo?

karianna:
@aahlenst merge conflict to resolve.

aahlenst:
I close this one for now. It's unclear whether we're going to use Artifactory in the future plus as far as I know no decision has been reached in the TSC whether we keep nightly builds forever or how long.

PR #130, opened 2019-06-18
libfontconfig1 as dependency or supply one?

nudgegoonies:
We have an application that needs to render fonts and does not run with adoptopenjdk until the libfontconfig1 package (debian) is installed. What would be the ideal solution?

Adding libfontconfig1 as package dependency for deb and rpm packages?

Add an own libfontconfig1 to adoptopenjdk tarballs like it is done with libfreetype?

I found out that the libfontconfig1 package needs libfreetype. I don't know how coupled they are. The adoptopenjdk tarballs and thus also the debian packages provide and use its own version of it. But the debian libfontconfig1 package depends on debians libfreetype package and thus adds a systemwide libfreetype library as well. When adoptopenjdk only uses libfreetype it uses its own one. When it uses libfontconfig then it uses the debian libfreetype from the debian package as well.


aahlenst:
Personally, I'd rather see the OS packages use the OS dependencies than adding yet another bundled dependency. As a consequence, I'd like to drop libfreetype, too.

I've asked whether this would be possible on the #build channel on Slack because I don't know the answer.

karianna:
What does Debian currently do in this space?

nudgegoonies:
@aahlenst
Thank you for the asking in the Channel. As i have written, i don't know how coupled they are. The don't have many transitive dependencies:
https://packages.debian.org/stretch/libfreetype6
https://packages.debian.org/stretch/libfontconfig1

@karianna 
Debian has a hard dependency to libfreetype6 AND libfontconfig even in the headless package:
https://packages.debian.org/stretch/openjdk-8-jre-headless
https://packages.debian.org/buster/openjdk-11-jre-headless
So that is no problem there. The headless package also depends on a lot of X11 packages.
The "headfull" package depends on the headless package and adds some more X11, font, audio, grahic library, etc. dependencies

aahlenst:
Seems related: https://github.com/AdoptOpenJDK/openjdk-build/issues/693. https://github.com/AdoptOpenJDK/openjdk-build/issues/693#issuecomment-458694375 suggests replacing the shipped libfreetype plus depending on libfontconfig could solve the issue and not trigger additional problems.

**But**: I'm not comfortable adding those changes blindly, though. I want tests that exercise the critical codepaths. I'm happy to incorporate those, just pointing me to some failing code would be enough. Running some multi-platform with the AdoptOpenJDK packages installed on the respective OS tests that cover installer-specific changes is on my to-do list anyway.

nudgegoonies:
I found a minimal example how to check font sizes with awt on stackoverflow and made a minimal executable of it:
```
import java.awt.Font;
import java.awt.font.FontRenderContext;
import java.awt.geom.AffineTransform;
class AWTestTSize {
public static void main(String[] args){
String text = "Hello World";
AffineTransform affinetransform = new AffineTransform();
FontRenderContext frc = new FontRenderContext(affinetransform,true,true);
Font font = new Font("Tahoma", Font.PLAIN, 12);
int textwidth = (int)(font.getStringBounds(text, frc).getWidth());
int textheight = (int)(font.getStringBounds(text, frc).getHeight());
}
}
```
With libfontconfig installed the program exits with 0 and without it exits with a NPE:
```
Exception in thread "main" java.lang.NullPointerException
	at sun.awt.FontConfiguration.getVersion(FontConfiguration.java:1264)
	at sun.awt.FontConfiguration.readFontConfigFile(FontConfiguration.java:219)
	at sun.awt.FontConfiguration.init(FontConfiguration.java:107)
	at sun.awt.X11FontManager.createFontConfiguration(X11FontManager.java:774)
	at sun.font.SunFontManager$2.run(SunFontManager.java:431)
	at java.security.AccessController.doPrivileged(Native Method)
	at sun.font.SunFontManager.<init>(SunFontManager.java:376)
	at sun.awt.FcFontManager.<init>(FcFontManager.java:35)
	at sun.awt.X11FontManager.<init>(X11FontManager.java:57)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at sun.font.FontManagerFactory$1.run(FontManagerFactory.java:83)
	at java.security.AccessController.doPrivileged(Native Method)
	at sun.font.FontManagerFactory.getInstance(FontManagerFactory.java:74)
	at java.awt.Font.getFont2D(Font.java:491)
	at java.awt.Font.access$000(Font.java:224)
	at java.awt.Font$FontAccessImpl.getFont2D(Font.java:228)
	at sun.font.FontUtilities.getFont2D(FontUtilities.java:180)
	at sun.font.StandardGlyphVector.initFontData(StandardGlyphVector.java:1126)
	at sun.font.StandardGlyphVector.init(StandardGlyphVector.java:1115)
	at sun.font.StandardGlyphVector.<init>(StandardGlyphVector.java:172)
	at java.awt.Font.getStringBounds(Font.java:2421)
	at java.awt.Font.getStringBounds(Font.java:2331)
	at AWTestTSize.main(AWTTestSize.java:12)
```

aahlenst:
Related: https://github.com/AdoptOpenJDK/openjdk-build/issues/682

aahlenst:
@nudgegoonies Thanks for the code.

For reference, I've checked Zulu 8. They don't bundle freetype with their tar.gz. And they specify freetype and fontconfig as required dependencies for their prebuilt packages.

nudgegoonies:
Any news about the supplied freetype? It would be great to have the dependency correct in the new 8u222, 11.0.4, 12.0.2 and 13.0.0 Versions.

aahlenst:
The fix should be ready for 13 (and at the same time for all other versions). Needs extensive testing and I haven‘t found the time yet. The plan is to strip freetype from all Linux builds and releases (even tarballs).

nudgegoonies:
Thank you for the answer. Very good. I know, that needs a lot of testing. But is this not part of the TCK tests during the build of the tarballs?

karianna:
> Thank you for the answer. Very good. I know, that needs a lot of testing. But is this not part of the TCK tests during the build of the tarballs?

We don't run TCK tests (bu do run a bunch of others)

aahlenst:
It depends on how early we can/want to remove the bundled freetype. The bundled freetype isn't an invention of AdoptOpenJDK but the upstream OpenJDK. Ideally, we replace freetype before running the test suite.

karianna:
FYI - It's unlikely we'll make this decision so close to the next quarterly release.

nudgegoonies:
Ok, thank you for that explantation.

nudgegoonies:
I will adjust the branch according to the decision by adding libfreetype to my MR or removing both libfreetype and libfontconfig.

mfriedenhagen:
Does setting the milestone to "October 2019" mean, we may expect this PR to be merged this month?

aahlenst:
No, that does not mean it. Martijn just does this every month.

It's not moving forward because I'm completely swamped with work. What's needed to make some progress:

1.  Take an empty machine (VM, container, whatever) without any dependencies, especially without libfontconfig and libfreetype, put our binary builds (not a Linux package) onto them and run the AdoptOpenJDK test suite on it. Are there any font-related test errors? If not, see that tests are added to the test suite that cover this because we currently have problems without libfontconfig.
2. On the same machine, remove the bundled libfreetype from binary builds. Install the distribution package for libfreetype, libfontconfig and run the test suite. Everything green? Then I'm happy to merge this pull request plus remove the bundled libfreetype. Otherwise we have to investigate further.

Should be tested at least with JDK 8, 11 and 13.

sxa555:
@willsparker The previous comment sounds like a good thing to try on a "clean" Vagrant VM :-)
@smlambert has recommended `BUILD_LIST=openjdk` with a `TARGET` of `jdk_awt` and/or `jdk_swing` as a reasonable place to start in terms of trying to run tests against it to expose any issues in this area.

aahlenst:
Superseded by #282.

PR #141, opened 2019-07-26
Improve integration of Linux package build with build pipelines

aahlenst:
During the last release, multiple problems surfaced:

* Mismatch of architecture names used by build pipelines and Gradle build caused build failures because not all Jenkins jobs have been updated.
* The Jenkins job that is part of the build pipeline pushed the releases into the repositories for the nightly builds because of misconfiguration.
* The Jenkins job that is part of the build pipeline failed if only a JDK but no JRE was available.
* The Jenkins job that actually pushes the releases frequently fails because it tries to re-release artifacts.

This pull request should address all these problems:

* Gradle build uses the same architecture signifiers as Jenkins build pipelines.
* Gradle build does not try to re-release existing artifacts, only updates metadata if necessary.
* Two shell scripts have been added that should replace the scripts in the Jenkins jobs. `create_installer_linux.sh` is for the regular build pipeline and creates installers from the tarballs copied into the Jenkins workspace. `standalone_create_installer_linux.sh` is a replacement for `artifactory_release_publisher` and `artifactory_release_publisher-bot`. As `standalone_create_installer_windows`, it downloads official releases from adoptopenjdk.net and repackages those using `create_installer_linux.sh`. Both shell scripts can cope with missing releases.


karianna:
@gdams and @johnoliver Can I get you to take a look - this will obviously be valuable if added with care.

@aahlenst Thanks for the PR - it looks sane to me although is there some tidy-up of old scripts and/or Jenkins jobs required as part of this?  Would be good to roughly document that.

aahlenst:
`artifactory_release_publisher-bot` and `artifactory_release_publisher` can be disabled and dropped after we have verified that the replacement is okay. `create_installer_linux.sh` should be a drop-in replacement for the job with the same name. The existing script that is part of the Jenkins job can be removed. Invoke `create_installer_linux.sh` instead. I can prepare and test the standalone job if someone would create it for me.

Does that suffice?

karianna:
> `artifactory_release_publisher-bot` and `artifactory_release_publisher` can be disabled and dropped after we have verified that the replacement is okay. `create_installer_linux.sh` should be a drop-in replacement for the job with the same name. The existing script that is part of the Jenkins job can be removed. Invoke `create_installer_linux.sh` instead. I can prepare and test the standalone job if someone would create it for me.
> 
> Does that suffice?

Sounds good to me.  Let me see what GitHub group you're part of and you can create your own job

karianna:
OK I'm blocked on a Jenkins 403 issue (we need to fix some Ngnix config) - I'll revisit this on Monday after we make that fix

aahlenst:
@karianna Could you already look into the Jenkins/nginx issue?

linux/build.gradle
```diff
@@ -86,8 +86,11 @@ def getJdkMajorVersion() {
         throw new IllegalArgumentException("Unparsable JDK major version '$versionString'")
     }
     int version = Integer.valueOf(versionString)
-    if (version < 8) {
-        throw new IllegalArgumentException("JDK major version '$version' out of range")
+
+    // Do not change the numbers without updating the metadata declarations
+    // (dependencies, jinfo priorities)!
+    if (version < 8 || version > 12) {
```
karianna: Should we be going to 13 now?

linux/buildSrc/src/main/java/net/adoptopenjdk/installer/UploadDebianPackage.java
```diff
@@ -46,10 +57,44 @@ protected void uploadPackages(ArtifactoryCredentials artifactoryCredentials) {
...
+        }
+
+        for (RepoPath searchResult : searchResults) {
+            if (Objects.equals(getRepository(), searchResult.getRepoKey())
+                    && Objects.equals(this.targetFilePath(), searchResult.getItemPath())) {
+                return true;
+            }
+        }
+
+        return false;
```
karianna: Should we log this case or is it common?

linux/buildSrc/src/main/java/net/adoptopenjdk/installer/UploadRpmPackage.java
```diff
@@ -50,14 +59,37 @@ protected void uploadPackages(ArtifactoryCredentials artifactoryCredentials) {
...
+                .artifactsByName(getPackageToPublish().getName())
+                .doSearch();
+
+        if (searchResults.isEmpty()) {
+            return false;
+        }
+
+        for (RepoPath searchResult : searchResults) {
+            if (Objects.equals(getRepository(), searchResult.getRepoKey())
+                    && Objects.equals(remotePath.substring(1), searchResult.getItemPath())) {
```
karianna: slight preference to pull `remotePath.substring(1)` into a named variable,  but no biggy.

linux/buildSrc/src/main/java/net/adoptopenjdk/installer/UploadRpmPackage.java
```diff
@@ -50,14 +59,37 @@ protected void uploadPackages(ArtifactoryCredentials artifactoryCredentials) {
...
+                return true;
             }
         }
+
+        return false;
```
karianna: Should we log this?

linux/create_installer_linux.sh
```diff
@@ -0,0 +1,135 @@
+#! /bin/bash
```
karianna: Has this whole file had shellcheck run on it?

linux/create_installer_linux.sh
```diff
@@ -0,0 +1,135 @@
...
+finish() {
+    if [ -d "$DISTRIBUTION_DIR" ] ; then
+        rm -rf "$DISTRIBUTION_DIR"
+    fi
+}
+trap finish EXIT
+
+echoerr() { printf "%s\n" "$*" >&2; }
+
+case $RELEASE_TYPE in
```
karianna: could this block be wrapped in a function?

linux/create_installer_linux.sh
```diff
@@ -0,0 +1,135 @@
...
+        ARTIFACTORY_REPOSITORY_RPM=""
+        GRADLE_TASK="build"
+        ;;
+    *)
+        echoerr "Unsupported build type"
+        exit 0
+        ;;
+esac
+
+JDK_TARBALLS=( "$WORKSPACE"/OpenJDK*.tar.gz )
```
karianna: could this block be wrapped in a function?

linux/build.gradle
aahlenst: It‘s prepared but comes in a separate pull request.

linux/buildSrc/src/main/java/net/adoptopenjdk/installer/UploadDebianPackage.java
aahlenst: It‘s logged, line 31.

linux/buildSrc/src/main/java/net/adoptopenjdk/installer/UploadRpmPackage.java
aahlenst: See line 36.

linux/create_installer_linux.sh
aahlenst: What would be the advantage? As far as I know, Bash does not really have return values and the code isn‘t re-used anywhere. Happy to make the change, but hard if I don‘t see what you are aiming at.

karianna: In that case, we can just add a comment at the top of the block explaining what it broadly does

karianna: Hmm, the previous comment didn't save.

In that case, perhaps add a comment at the top of the block?

linux/create_installer_linux.sh
aahlenst: Yes, but I missed one issue. It's fixed now.

linux/create_installer_linux.sh
aahlenst: Comments have been added.

Issue #137, opened 2019-07-16
jvm.dll location has changed for Windows X86-32 Hotspot from bin/server to bin/client


douph1:
This change is not yet taking into installer

PS: no change for openj9

This change was found by looking at pre-release 8u222 internal zip

douph1:
I'm looking at the commit in https://github.com/AdoptOpenJDK/openjdk-build wich trigger this change, is someone can help me to find it to confirm.
I don't think it is documented somewhere in a changelog

karianna:
@ali-ince will know

ali-ince:
Hi @douph1,

The specific change [#1104](https://github.com/AdoptOpenJDK/openjdk-build/pull/1104), [#1113](https://github.com/AdoptOpenJDK/openjdk-build/pull/1113) and [#1114](https://github.com/AdoptOpenJDK/openjdk-build/pull/1114) are made to get `client` and `server` jvm variants into the distribution whereever it's possible. 

The change was initially done to get x86 hotspot binaries in line with other builds (due to [#871](https://github.com/AdoptOpenJDK/openjdk-build/issues/871)) but later on extended to include both x86 and x64 binaries (except for JDK8 x64 where `client` variant is not supported).

The change has the following consequences;

| Version | Configuration | Included Variants |
| :--- | :--- | :--- |
| Java 8 | JDK x86 | `jre/bin/client/jvm.dll` and `jre/bin/server/jvm.dll`  | 
| | JRE x86 | `bin/client/jvm.dll`  |
| | JDK x64 | `jre/bin/server/jvm.dll`  |
| | JRE x64 | `bin/server/jvm.dll`  |
| Java 11 | JDK x86 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  | 
| | JRE x86 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  |
| | JDK x64 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  |
| | JRE x64 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  |
| Java 12 | JDK x86 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  | 
| | JRE x86 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  |
| | JDK x64 | `bin/client/jvm.dll` and `bin/server/jvm.dll`  |
[...]

PR #142, opened 2019-07-26
fix build failed for JDK since last change about RuntimeLib jvm.dll

douph1:
fix for build failed after PR for #137


douph1:
ready for merge

PR #143, opened 2019-07-26
Changelog for Windows Installer

douph1:
As discuted @gdams

### DIFF
--- a/wix/InstallerChangelog.md
+++ b/wix/InstallerChangelog.md
@@ -1,0 +1,82 @@
+# Changelog for AdoptOpenJDK Windows installer
+
+# Changelog
+* All notable changes **impacting the users** will be documented in this file.
+* Technical changes/build fix/rewrite will not be covered here for readability if the produced msi is not impacted.
+* For more/technical changes or previous changes see [Commit History](https://github.com/AdoptOpenJDK/openjdk-installer/commits/master) 
+
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
+
+## 2021-01-25
+### Added
+- This user oriented changelog is born [#143](https://github.com/AdoptOpenJDK/openjdk-installer/issues/143)
+
+## 2021-01-25
+### Added
+- Windows Installer can package Alibaba Dragonwell [#280](https://github.com/AdoptOpenJDK/openjdk-installer/issues/280)
+
+## 2020-12-16
+### Added
+- Windows Installer can package aarch64/arm64 JDK [#277](https://github.com/AdoptOpenJDK/openjdk-installer/issues/277)
+
+## 2020-11-19
+### Cleanup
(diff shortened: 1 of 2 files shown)


douph1:
I have rebased this near 1year old PR ..
If interested in tracked/publish the Windows Installer change to community, please take a look and comment or reject/drop

aahlenst:
Having a changelog is a great thing. But it does not make sense to do that only for the MSIs and not by date. The changes should either be grouped by JDK versions or by CPU dates. An "Upcoming" section is needed, too.

What was the outcome of your discussion with George?

douph1:
- Me: What are you thinking about starting to maintain a "Changelog" for the Windows installer ?

- G: yeah that would be a really good idea! do you want to put something together?

- Me: Did we must try to recover the past history to document it ? (please say no :) )

  I just don't know how tagged the change as we don't version the installer .. there is only a master 
  branch wich is used in release
  so matchin feature add/remove with release of 8u222 or something else is difficult
 so maybe just set the changelog by date/time .. and if 8u222b12 b18 b25 is build after this date we can expect the new feature is embeded in the patch release ..

- G: lets go with no
 start from now
 and yeah do it by date

- M: last question :), do you think we must embeded the changelog in the distributed JDK directory wich is installed on the client computer ? 

- G: no I'd just keep it in github 

- Me: k

- G: unless you think embedding it makes sense

- Me: a little but i can make confusion with Changelog from OpenJDK
 Or dont naming it ChangeLog but InstallerChangelog ...
 I think it can be helpful to know what the installer is capable

- G: InstallerChangelog is probably best

- Me: if you setup Adop msi .. next next next , you go to folder and you know what is the embeded fix into this msi

- G: sounds good to me!

- Me: Ok I will try to do a draft for that later

- G: great!
thanks for all your work on the installers!

aahlenst:
I think it makes sense to have a changelog, but it does not make sense to have one without proper releases. There's no much difference to the Git history. The problem here seems to be that we always build from master instead of tags. This also requires us to stop the development during the releases.

douph1:
So we must start a new issue to discuss what change are required to build from tagged openjdk-installer ? right ?
How tag it ? Manually ( In sync with code freeze/ cpu date / release ) ?

aahlenst:
I guess some discussion is going to be needed. Ideally, this effort should extend to other repositories. Apart from that, we might need to split the installer repository for Mac, Windows and Linux. This would also make sense from a testing POV. Triggering multiple hours of tests for a installer variant that hasn't changed isn't a good use of our computing resources. I've started this process for Adoptium/Linux.

karianna:
@douph1 - Can you update this PR with the various release notes and then I think this can go in :-)

gdams (approved):

douph1:
I have updated the changelog and added some clarification about the goal wich is not to duplicate commit history but grouping visible change impacting the consumer of the MSI

PR #153, opened 2019-08-19
Validate VERSION parameter

aahlenst:
We should validate the contents of the VERSION argument. Typos or mismatches cause changes to the package names and as a result, we might push releases to Artifactory with an invalid name (invalid as in "might break updates"). In the worst case, we'd have to pull a release.

Could someone please take an extra look at the patterns, especially those involving OpenJ9? Do they match the versions used by the build pipelines?

### DIFF
--- a/linux/create_installer_linux.sh
+++ b/linux/create_installer_linux.sh
@@ -124,6 +124,66 @@ for JDK_TARBALL in ${JDK_TARBALLS[*]} ; do
             ;;
     esac
 
+    # Checks the format of the VERSION variable. A consistent formatting is
+    # crucial because accidental format changes might break package updates.
+    case $RELEASE_TYPE in
+        "Release")
+            if [ "$MAJOR_VERSION" -eq 8 ] ; then
+                if [ "$JVM" == "hotspot" ] ; then
+                    # Should look like: 8u222-b10
+                    VERSION_PATTERN='8u[0-9]+\-b[0-9]+'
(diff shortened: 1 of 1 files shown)


linux/create_installer_linux.sh
```diff
@@ -124,6 +124,66 @@ for JDK_TARBALL in ${JDK_TARBALLS[*]} ; do
             ;;
     esac
 
+    # Checks the format of the VERSION variable. A consistent formatting is
```
karianna: @johnoliver and @gdams - we have a shared lib for version parsing now don't we?  Please tell me we do :-D

aahlenst: I just found https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/library/src/common/VersionInfo.groovy which doesn't do what's needed here. I'm mainly interested in checking that the combination of the version and release type is legit.

johnoliver: We do have a version parsing library (what @aahlenst linked to), but it probably would need to be enhanced in order to deal with some of this uses case, i.e add some sort of validate function, and probably parsing of the openj9 portions of the version string. But that is probably something we should do

johnoliver: may be leaning towards accepting this unless we have time to upgrade the groovy lib some time soon

karianna: OK apologies for the delay

PR #175, opened 2019-11-01
Add support for Fedora 31, drop Fedora 29

aahlenst:
Plus the addition of JDK 13 to the support matrix that was previously undocumented because I forgot to do it.


johnp:
The rpms for Fedora 31 are still [missing](https://adoptopenjdk.jfrog.io/adoptopenjdk/rpm/fedora/). Is there are timeline as to when they are built?

aahlenst:
Waiting: https://github.com/AdoptOpenJDK/TSC/issues/127

PR #176, opened 2019-11-01
Add support to disable macOS signing

junyzheng:
Build the macOS installer when `CERTIFICATE` is not set. User can add their own customized steps to sign the macOS package.

### DIFF
--- a/pkgbuild/create-installer-mac.sh
+++ b/pkgbuild/create-installer-mac.sh
@@ -27,8 +27,15 @@ if [ ! -z "$CERTIFICATE" ]; then
 fi
 set -u
 
+set +u
+SEARCH_PATTERN=
+if [ -z "$SEARCH_PATTERN" ]; then
+  SEARCH_PATTERN=OpenJDK*-j*.tar.gz
+fi
+set -u
+
 cd pkgbuild
-for f in $WORKSPACE/workspace/target/*.tar.gz;
+for f in $WORKSPACE/workspace/target/${SEARCH_PATTERN};
 do tar -xf "$f";
   rm -rf Resources/license.rtf
 
--- a/pkgbuild/create-installer-mac.sh
+++ b/pkgbuild/create-installer-mac.sh
@@ -20,8 +20,15 @@ if [ -f ~/.password ]; then
   security unlock-keychain -p `cat ~/.password`
 fi
 
+set +u
+SIGN_OPTION=
+if [ ! -z "$CERTIFICATE" ]; then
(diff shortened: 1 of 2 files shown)


junyzheng:
@gdams do you mind to review this? thank you!

pkgbuild/create-installer-mac.sh
```diff
@@ -20,8 +20,15 @@ if [ -f ~/.password ]; then
...
+set +u
+SIGN_OPTION=
+if [ ! -z "$CERTIFICATE" ]; then
+  SIGN_OPTION="--sign ${CERTIFICATE}"
+fi
+set -u
+
 cd pkgbuild
-for f in $WORKSPACE/workspace/target/OpenJDK*-j*.tar.gz;
+for f in $WORKSPACE/workspace/target/*.tar.gz;
```
gdams: unfortunately, this will break the job, we had an issue where  it tried to create an installer for the `testimage` as well. Maybe you can add a flag for this or something?

pkgbuild/pkgbuild.sh
```diff
@@ -122,20 +124,22 @@ cat distribution.xml.tmpl  \
...
-echo "Notarizing the installer (please be patient! this takes aprox 10 minutes)"
-sudo xcode-select --switch /Applications/Xcode.app || true
-cd notarize
-npm install
-node notarize.js --appBundleId $IDENTIFIER --appPath ${OUTPUT_DIRECTORY}
-# Validates that the app has been notarized
-spctl -a -v --type install ${OUTPUT_DIRECTORY}
-cd -
-#fi
+if [ ! -z "$SIGN_OPTION" ]; then
```
gdams: unhappy with this, you probably need to introduce a `NOTARIZE_OPTION` or something, the two processes are not the same.

gdams:
reviewed

pkgbuild/create-installer-mac.sh
junyzheng: OK, I will add an option.

pkgbuild/pkgbuild.sh
junyzheng: OK, I will create another internal option: `NOTARIZE_OPTION`.

gdams:
hmm @junyzheng we're now seeing this error:

```
./pkgbuild.sh: line 67: INPUT_DIRECTORY: unbound variable
```

Any idea what needs fixing?

junyzheng:
hmmm... OK, let me take a look.

gdams:
okay think I've fixed it now :)

PR #177, opened 2019-11-05
Fix issue in PR #176

junyzheng:
@gdams 

I tested in Microsoft Juniper Pipeline: `https://dev.azure.com/ms-juniper/Juniper/_build/results?buildId=2144&view=logs&j=438df1b3-061f-559b-4ceb-8133ae4cf5b7`. I also add a `echo` command to improve the debug ability. I think the issue you saw `INPUT_DIRECTORY: unbound variable` is probably related to the script cannot find the file.

### DIFF
--- a/pkgbuild/create-installer-mac.sh
+++ b/pkgbuild/create-installer-mac.sh
@@ -28,7 +28,6 @@ fi
 set -u
 
 set +u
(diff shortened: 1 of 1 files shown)


karianna:
@gdams you'll want to check the result of this

PR #208, opened 2020-05-12
Hardening detection of jdk/jvm/arch in filename

### LINKED ISSUE #207: Windows installer build failling if Jenkins artifact name is not as expected
./CreateSourceFolder.AdoptOpenJDK.ps1 look zip filename to find which jdk/jvm/arch version this artifact is.

The job https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk/job/jdk-windows-x64-hotspot/
always failed because the artifact zip filename doesn't contain the JDK major version as usualy.

douph1:
- exit with error code and message if regex don't match
 - Zip must begin with "^" added
 Fix #207


aahlenst:
Thanks!

PR #210, opened 2020-05-19
Account for JDK-Next Binaries

M-Davies:
* JDK-NEXT tarballs have no major version number in name

Closes: https://github.com/AdoptOpenJDK/openjdk-build/issues/1712
Signed-off-by: Morgan Davies <morgan.davies@ibm.com>

### DIFF
--- a/wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
+++ b/wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
@@ -1,6 +1,6 @@
 Get-ChildItem -Path .\ -Filter *.zip -File -Name| ForEach-Object {
   $filename = [System.IO.Path]::GetFileName($_)
-  $jdk_version_found = $filename -match "(?<jdk>^OpenJDK\d+)"
+  $jdk_version_found = $filename -match "(?<jdk>^(?:OpenJDK\d+|OpenJDK))"
   if (!$jdk_version_found) {
     echo "filename : $filename don't match regex ^OpenJDK\d+"
     exit 2


douph1:
Hum this fix is strange .. we now accept OpenJDK without knowing the major version.
The variable jdk_version_found is now true .. but we may have not found a jdk major version
and last, the error message wich display regex no longer match the regex tested ..

douph1:
Also the PR was not tested in any other job.
Usualy we test the PR with test_installer_windows or standalone_create_installer_windows job before merging.
Then, this PR didn't fix the bug as the extracted folder don't match what expected
https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/create_installer_windows/4110/console
We are looking against 3 known extract path :
```
First .\SourceDir\OpenJDK15\hotspot\x64\jdk-15.0.0 not exists
Second .\SourceDir\OpenJDK15\hotspot\x64\jdk-15+24 not exists
Third .\SourceDir\OpenJDK15\hotspot\x64\jdk-15.0.0+24 not exists
```
but the zip file is now extracted ... without the major version determined
`SourceDir\OpenJDK\hotspot\x64\jdk-15+24`

As discussed in slack we need to stop extract all zip file blindly
`Get-ChildItem -Path .\ -Filter *.zip -File -Name| ForEach-Object`
and trying to detect arch/jvm/major for each but just rely on the jobs args and manage one artifact at a time.

We need another extract script. I will make a new bug to address this cleanly

douph1:
see https://github.com/AdoptOpenJDK/openjdk-installer/issues/211
I tried to summarize why we have this problem and all that it entails

PR #214, opened 2020-05-29
Add RISCV support

sxa:
This will likely fail at present due to the build pipelines being `riscv` instead of `riscv64` but that is being addressed under a separate PR.

Fixes https://github.com/AdoptOpenJDK/openjdk-installer/issues/213

At the moment I don't have a Debian system to test on, and I suspect there is no SuSE port to this platform so that is superfluous.

I have verified that the RedHat rpm produced by this will install and is runnable on my Fedora RISCV system, however it does not uninstall so further work will be required:

```
[root@stage4 /]# rpm -e adoptopenjdk-11-openj9-jre-11.0.0_internal+0_adhoc.jenkins.openj9_openjdk_jdk11-1.riscv64
error: %preun(adoptopenjdk-11-openj9-jre-11.0.0_internal+0_adhoc.jenkins.openj9_openjdk_jdk11-1.riscv64) scriptlet failed, exit status 2
error: adoptopenjdk-11-openj9-jre-11.0.0_internal+0_adhoc.jenkins.openj9_openjdk_jdk11-1.riscv64: erase failed
[root@stage4 /]# 
```

Signed-off-by: Stewart Addison <sxa@uk.ibm.com>

### DIFF
--- a/linux/buildSrc/src/main/java/net/adoptopenjdk/installer/Architecture.java
+++ b/linux/buildSrc/src/main/java/net/adoptopenjdk/installer/Architecture.java
@@ -9,6 +9,7 @@ package net.adoptopenjdk.installer;
 public enum Architecture {
     X64("amd64", "x86_64", 64),
     S390X("s390x", "s390x", 64),
+    RISCV64("riscv64", "riscv64", 64),
     PPC64LE("ppc64el", "ppc64le", 64),
     ARM("armhf", "armhfp", 32),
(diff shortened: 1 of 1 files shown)


sxa:
I think the error I'm getting is because it's trying to remove a `javac` alternative that it hasn't put in place because the package thinks it's a JRE, not a JDK

sxa:
The same failure occurs on Linux/x64, therefore assuming that it will identify properly as a JDK this change should be good - removing from draft

PR #218, opened 2020-06-05
fix bad syntax

douph1:


douph1:
this is to fix new error since https://github.com/AdoptOpenJDK/openjdk-installer/pull/217
`call Build.OpenJDK_generic.cmd 
'ADOPT_NAME' is not recognized as an internal or external command,
operable program or batch file.`

PR #221, opened 2020-06-10
allow any vendor to produce packages with vendor metadata and name

jennli:
[UPDATED]
Will keep the same gradle task but add optional arguments to specify VENDOR, VENDOR_HOMEPAGE, VENDOR_SROURCE_URL, DEBIAN_ITERATION

**[DEPRECATED INFO BELOW]**
Proposing a new gradle task for building .deb package by any vendor

*VENDOR*:
- vendor value is passed in by providing `-PVENDOR="vendor.."` to the gradle task
- if not specified, it defaults to "AdoptOpenJDK". 

*VENDOR_HOMEPAGE*:
- homepage is passed in by providing `-PVENDOR_HOMEPAGE="www.abc.com"` to the gradle task.
- if not specified, the value defaults to "https://adoptopenjdk.net/"

*Testing Results*:
An example package is created by `./gradlew buildVendorDebPackage` with the following arguments: 
```
-PJDK_DISTRIBUTION_TYPE=JDK 
-PJDK_MAJOR_VERSION=11 
-PJDK_VERSION=11.0.8+2 
-PJDK_VM=hotspot 
-PVENDOR=Microsoft 
-PVENDOR_HOMEPAGE="www.microsoft.com" 
...
```
results in a package name of `openjdk-11-hotspot-11.0.8+2_Microsoft_amd64.deb`
and this `dpkg -I` output:
```
 new Debian package, version 2.0.
 Package: openjdk-11-hotspot
 Version: 11.0.8+2-2
 License: GPL-2.0+CE
 Vendor: Microsoft
 Architecture: amd64
 Maintainer: Microsoft
 Installed-Size: 369769
 Depends: ...
 Provides: ...
 Homepage: https://www.microsoft.com/
 Description: OpenJDK Development Kit 11 (JDK) with Hotspot by Microsoft
```

*Testing that my change is not breaking Adopt*
a package created by `./gradlew buildDebPackage` task with no vendor or vendor homepage specified, results in the pacakge name of `adoptopenjdk-11-hotspot_11.0.8+2-2_amd64.deb`
and the following`dpkg -I` output:
```
 new Debian package, version 2.0.
 Package: adoptopenjdk-11-hotspot
 Version: 11.0.8+2-2
 Vendor: AdoptOpenJDK
 Architecture: amd64
 Maintainer: AdoptOpenJDK
 Installed-Size: 369769
 Homepage: https://adoptopenjdk.net/
 Description: OpenJDK Development Kit 11 (JDK) with Hotspot by AdoptOpenJDK
...
```

### DIFF
--- a/linux/README.md
+++ b/linux/README.md
@@ -54,10 +54,44 @@ Example:
 ```
 
 #### Optional Arguments
-- `-PVENDOR=vendor_name` - Specify a custom vendor name
-- `-PVENDOR_HOMEPAGE="www.homepage.com"` - specify a custom link to vendor homepage
-- `-PVENDOR_SOURCE_URL="www.sourcecode.com"` - specify a custom link to source code
-- `-PDEBIAN_ITERATION=1` - specify the iteration
+
+* `-PPACKAGE_NAME=pkg_name` - Specify the name of the output package, defaults to `adoptopenjdk`
+* `-PVENDOR=vendor_name` - Specify a custom vendor name, defaults to `AdoptOpenJDK`
+* `-PVENDOR_HOMEPAGE="https://example.com/"` - specify a custom link to vendor homepage
+* `-PVENDOR_SOURCE_URL="https://example.com/"` - specify a custom link to source code
+* `-PDEBIAN_ITERATION=1` - specify the iteration
+
+3rd party vendors can choose to populate the above optional arguments to customize package metadata.
+
+Example:
+
+```bash
+./gradlew buildDebPackage \
+    -PJDK_DISTRIBUTION_TYPE=JDK \
+    -PPACKAGE_NAME=openjdk \
+    -PVENDOR=Vendor \
+    -PVENDOR_HOMEPAGE="https://homepage.com" \
+    -PJDK_DISTRIBUTION_DIR=path-to/jdk-11.0.8+2 \
+    -PJDK_MAJOR_VERSION=11 \
+    -PJDK_VERSION=11.0.8+2~vendor \
+    -PJDK_VM=hotspot \
+    -PJDK_ARCHITECTURE=x64 \
+    -PDEBIAN_ITERATION=5
+```
+
+will produce `openjdk-11-hotspot_11.0.8+2~vendor-5_amd64.deb` with the following metadata:
+
+```
+ new Debian package, version 2.0.
+ Package: openjdk-11-hotspot
+ Version: 11.0.8+2~vendor-5
+ Vendor: Vendor
+ Maintainer: Vendor
+ Homepage: https://homepage.com
+ Description: OpenJDK Development Kit 11 (JDK) with Hotspot by Vendor
+ ...
+
+```
 
 Table with arguments:
 
--- a/linux/README.md
+++ b/linux/README.md
@@ -53,6 +53,12 @@ Example:
     -PJDK_ARCHITECTURE=x64
 ```
 
+#### Optional Arguments
+- `-PVENDOR=vendor_name` - Specify a custom vendor name
+- `-PVENDOR_HOMEPAGE="www.homepage.com"` - specify a custom link to vendor homepage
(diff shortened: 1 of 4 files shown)


aahlenst:
Why did you opt to create a separate task? I'd rather make the package name configurable. In that case, we do not have to worry about components using "openjdk" in their name but containing OpenJ9.

aahlenst:
Before making further changes, let us talk about the naming requirements first. As soon as we have something in place, we should be able to land this quickly.

Context:
* `openjdk-...`, and `java-...` respectively, are only used by Linux distributors. All OpenJDK vendors use a custom name: `adoptopenjdk-...`, `zulu-...` etc.
* A vendor's name isn't something that is usually part of the package name and would not appear in the output of `apt search`. It does not make sense from a distributor's POV because usually all packages are made by them. I've seen vendor names only in file names, e.g. `openjdk-11-jdk-headless_11.0.7+10-3~deb10u1_amd64.deb` or `openjdk-11-jdk-headless_11.0.7+10-2ubuntu2~19.10_amd64.deb`. As you can see, both Debian and Ubuntu treat it as part of the version tag.

BTW: Using an underline is problematic because AFAIK it's treated as a separator by Debian/Ubuntu.

jennli:
Thanks @aahlenst , I have updated my code, please take a look.

linux/build.gradle
```diff
@@ -74,6 +82,22 @@ def getJdkArchitecture() {
         Architecture.X64
 }
 
+def getVendor() {
+    return hasProperty("VENDOR") ? VENDOR : "AdoptOpenJDK"
+}
+
+def getPackageName() {
+    return hasProperty("VENDOR") ? "openjdk" : "adoptopenjdk"
```
aahlenst, line 90: I have doubts that every third party wants to call their package openjdk. I'd introduce a separate property `PACKAGE_NAME`.

linux/README.md
```diff
@@ -53,6 +53,12 @@ Example:
     -PJDK_ARCHITECTURE=x64
 ```
 
+#### Optional Arguments
+- `-PVENDOR=vendor_name` - Specify a custom vendor name
+- `-PVENDOR_HOMEPAGE="www.homepage.com"` - specify a custom link to vendor homepage
```
aahlenst, line 58: Reading the code, the homepage has to include the protocol. Furthermore, example.com is reserved for this purpose: `-PVENDOR_HOMEPAGE="https://example.com/"`

linux/README.md
```diff
@@ -53,6 +53,12 @@ Example:
...
+- `-PVENDOR_SOURCE_URL="www.sourcecode.com"` - specify a custom link to source code
```
aahlenst, line 59: Protocol here, too.

linux/deb/build.gradle
```diff
@@ -13,9 +13,9 @@ ext.jinfoPriorities = [
 ]
 
 tasks.register("buildDebianPackage", BuildDebianPackage) {
-    packageName = "adoptopenjdk-${jdkMajorVersion}-${pkgMetadata.vm.packageQualifier}${jdkDistributionType.pkgNameSuffix}"
+    packageName = pkgMetadata.packageName
```
aahlenst, line 16: Please do that for the RPMs, too.

linux/README.md
```diff
@@ -53,6 +53,12 @@ Example:
     -PJDK_ARCHITECTURE=x64
 ```
 
+#### Optional Arguments
```
aahlenst, line 56: Ideally, this section should explain how third parties can make packages with their own naming scheme. It should also mention that company names are supposed to go into the version.

linux/README.md
jennli: good to know 💡

linux/deb/build.gradle
jennli: added to both rpm tasks

PR #241, opened 2020-07-31
Allow installers to build in pipelines without artifactory

sxa:
This allows the installers to be built but not uploaded to artifactory (by forcing `GRADLE_TASK=build`) if artifactory credentials have not been supplied, for the situation where someone outside adoptopenjdk is running our pipelines (since this will currently fail in that situation and not allow the installers to be archived)

We should possibly also look at making the artifactory server name a parameter along with the username and password to incfrease flexibility

Signed-off-by: Stewart Addison <sxa@uk.ibm.com>

### DIFF
--- a/linux/create_installer_linux.sh
+++ b/linux/create_installer_linux.sh
@@ -59,7 +59,7 @@ echoerr() { printf "%s\n" "$*" >&2; }
 # ARTIFACTORY_REPOSITORY_DEB or ARTIFACTORY_REPOSITORY_RPM. If it's a nightly
 # build without publication, instruct Gradle only to build the packages
 # without uploading them.
-case $RELEASE_TYPE in
+case "$RELEASE_TYPE" in
     "Release")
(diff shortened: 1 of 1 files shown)


aahlenst:
@karianna The future refactoring should look like this: Remove publishing from Gradle and delegate it to Jenkins. The objective is to let the Jenkins master collect the packages and SFTP them wherever we want them to end up. Furthermore, if we relied on Gradle properties in the future, they should be lazy so that they've to be defined only when required.

Generally, the `*_create_installer_linux.sh` scripts are glue for Adopt's Jenkins and aren't supposed to be consumed by third parties. But I don't object to the PR.

sxa:
> Generally, the `*_create_installer_linux.sh` scripts are glue for Adopt's Jenkins and aren't supposed to be consumed by third parties. But I don't object to the PR.

Unfortunately the main build pipelines require them so it's not possible to consume them as-is without an installer job being available

PR #243, opened 2020-08-23
Update list of alternatives

### LINKED ISSUE #242: Check whether alternatives in Linux packages are up-to-date
I noticed today that `jpackage` does not have an alternative in AdoptOpenJDK 14. Probably `jfr` is missing somewhere, too. So we have to go through 8, 11 and 15 and check whether all executables have an alternative configured.

aahlenst:
Fixes #242


karianna (approved):
LGTM

PR #244, opened 2020-08-26
remove old p12 codesign cert

gdams:


aahlenst (changes requested):
Can I get a sentence like "Because we store it in the secrets repo for security reasons and place it on the machines with Ansible" in the commit description?

gdams:
@aahlenst updated PTAL

aahlenst (approved):

PR #253, opened 2020-10-21
Replace SIGN_OPTION with SIGN_CERT

AdamBrousseau:
I assume this was an oversight in the original authorship.
If CERTIFICATE is not defined the script will fail because
SIGN_CERT is unbound.

Signed-off-by: Adam Brousseau <adam.brousseau@ca.ibm.com>


AdamBrousseau:
@gdams Can I get a review please?

gdams (approved):

PR #257, opened 2020-11-06
Move quotes around SIGN_CERT into variable itself

### LINKED ISSUE #256: Mac installer packagebuild.sh unset variable INPUT_DIRECTORY when Certificate is not used.
For the use case where we want to create a mac installer and do not use a certificate. If we don't pass certificate, the package command will end up being, for example
```
./packagesbuild.sh '' --major_version 8 --full_version 1.8.0_272-202011032038-b10 --input_directory jdk8u272-b10 --output_directory /Users/jenkins/workspace/build-scripts/release/create_installer_mac/workspace/target/OpenJDK8U-jdk_x64_mac_openj9_2020-11-03-20-32.pkg --jvm openj9 --type jdk
```
Note the two single quotes that would normally wrap SIGN_CERT. This causes the script to fail
```
./packagesbuild.sh: line 108: INPUT_DIRECTORY: unbound variable
```
Rerunning the command manually without the quotes works.
If I remove the `set -u` I can see what is/isn't set
```
MAJOR_VERSION:8
FULL_VERSION:1.8.0_272-202011032038-b10
INPUT_DIRECTORY:
OUTPUT_DIRECTORY:
JVM:openj9
TYPE:
VENDOR:adoptopenjdk
PACKAGE_NAME:AdoptOpenJDK
LOGO:Resources/adoptopenjdk.png
IDENTIFIER:
SIGN_OPTION:
SIGN_CERT:
NOTARIZE_OPTION:
```
From what I can deduce:
MAJOR_VERSION, FULL_VERSION are set because they are parameters
JVM is set because it gets exported
VENDOR, PACKAGE_NAME, LOGO are all ~set~ hardcoded in packagebuild.sh, which is another issue.
IDENTIFIER, SIGN_OPTION, SIGN_CERT, NOTARIZE_OPTION not set as expected.

Can the quotes for SIGN_CERT be moved into the variable?
It appears maybe yes. I will open a PR.

AdamBrousseau:
When CERTIFICATE is not used, the empty quotes around
SIGN_CERT in the packagebuild command will mess up the
input args. Moving the quotes into the variable will
remove them from the call when the variable doesn't get set.

Fixes #256
Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

### DIFF
--- a/pkgbuild/create-installer-mac.sh
+++ b/pkgbuild/create-installer-mac.sh
@@ -21,11 +21,12 @@ if [ -f ~/.password ]; then
 fi
 
 set +u
-SIGN_CERT=
-SIGN_CMD=
-if [ ! -z "$CERTIFICATE" ]; then
+if [ -z "$CERTIFICATE" ]; then
+  SIGN_CMD=
+  SIGN_CERT=
+else
   SIGN_CMD="--sign"
-  SIGN_CERT="${CERTIFICATE}"
+  SIGN_CERT="'${CERTIFICATE}'"
 fi
 set -u
 
@@ -64,8 +65,8 @@ do tar -xf "$f";
   directory=$(ls -d jdk*)
   file=${f%%.tar.gz*}
 
-  echo running "./packagesbuild.sh ${SIGN_CMD} "${SIGN_CERT}" --major_version ${MAJOR_VERSION} --full_version ${FULL_VERSION} --input_directory ${directory} --output_directory ${file}.pkg --jvm ${JVM} --type ${TYPE}"
-  ./packagesbuild.sh ${SIGN_CMD} "${SIGN_CERT}" --major_version ${MAJOR_VERSION} --full_version ${FULL_VERSION} --input_directory ${directory} --output_directory ${file}.pkg --jvm ${JVM} --type ${TYPE}
+  echo running "./packagesbuild.sh ${SIGN_CMD} ${SIGN_CERT} --major_version ${MAJOR_VERSION} --full_version ${FULL_VERSION} --input_directory ${directory} --output_directory ${file}.pkg --jvm ${JVM} --type ${TYPE}"
+  ./packagesbuild.sh ${SIGN_CMD} ${SIGN_CERT} --major_version ${MAJOR_VERSION} --full_version ${FULL_VERSION} --input_directory ${directory} --output_directory ${file}.pkg --jvm ${JVM} --type ${TYPE}
 
   rm -rf ${directory}
   rm -rf ${f}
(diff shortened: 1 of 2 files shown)


AdamBrousseau:
I was able to test and verify this works for my use case. I would recommend we test this at Adopt for the typical use case.

karianna:
Does {} also capture the "" use case?

AdamBrousseau:
Hi @karianna , I'm not sure I understand what change you're suggesting?

karianna:
> Hi @karianna , I'm not sure I understand what change you're suggesting?

Just making sure "${CERTIFICATE}" is as safe (escaping wise) as "'${CERTIFICATE}'"

pkgbuild/create-installer-mac.sh
```diff
@@ -25,7 +25,7 @@ SIGN_CERT=
 SIGN_CMD=
 if [ ! -z "$CERTIFICATE" ]; then
   SIGN_CMD="--sign"
-  SIGN_CERT="${CERTIFICATE}"
+  SIGN_CERT="'${CERTIFICATE}'"
```
keithc-ca, line 28: I think it reads better without the negation, assigning both variables exactly once:
```
if [ -z "$CERTIFICATE" ] ; then
  SIGN_CMD=
  SIGN_CERT=
else
  SIGN_CMD="--sign"
  SIGN_CERT="'${CERTIFICATE}'"
fi
```

keithc-ca:
> Just making sure "${CERTIFICATE}" is as safe (escaping wise) as "'${CERTIFICATE}'"

Yes, I think it will work as intended; we'll either get
```
  packagesbuild.sh   --major_version ...
```
or
```
packagesbuild.sh --sign 'certificate name' --major_version ...
```

pkgbuild/create-installer-mac.sh
AdamBrousseau: Good idea. It is safer to set something to blank rather than assume it is not set.

karianna (approved):
LGTM

karianna:
Will have to merge post release

pkgbuild/create-installer-mac.sh
```diff
@@ -23,9 +23,12 @@ fi
 set +u
 SIGN_CERT=
 SIGN_CMD=
```
keithc-ca, line 25: These lines should be deleted (see lines 27-28).

AdamBrousseau: good catch, updated.

gdams (approved):

sxa:
> I was able to test and verify this works for my use case. I would recommend we test this at Adopt for the typical use case.

Was this tested at adopt prior to merging? It looks like it's caused breakages on all the macos install jobs today - possible because it's starting by passing in `''` as the first parameter without `--sign`

gdams:
@sxa I created a github action in this repo to test the change. It looks like @andrew-m-leonard's fix is good so I'll merge

sxa:
@gdams You mean a new action?

gdams:
@sxa https://github.com/AdoptOpenJDK/openjdk-installer/pull/268

sxa:
OK so Adam's PR wasn't tested prior to merging

PR #266, opened 2020-11-20
Add full version number to PKG metadata

### LINKED ISSUE #263: Version number reverted on Munki import
When importing the pkg into Munki it determines the version  from the pkg file. 

EG
OpenJDK 8u262b10 | 1.8.0_262-b10 | macOS | Custom | Munki
OpenJDK8U-jdk_x64_mac_hotspot_8u | 1.8.0_265-b01 | macOS | Custom | Munki

However after the 265-b01 version, the version number is now 1.0

OpenJDK8U-jdk_x64_mac_hotspot_8u272b10 | 1.0 | macOS | Custom | Munki

This then prevents munki from pushing out the update. Is this able to be reverted back to including the full version description?

munki_jdk

aahlenst:
During the migration to Packages, we stopped adding the full version number to the PKG metadata. Instead, it was hardcoded to 1.0.

Additionally, this PR fixes inconsistent capitalization of the file OpenJDKPkg.pkgproj.template which is an issue on Macs with a case-sensitive file system.

Fixes #263.

### DIFF
--- a/pkgbuild/OpenJDKPKG.pkgproj.template
+++ b/pkgbuild/OpenJDKPKG.pkgproj.template
@@ -164,7 +164,7 @@
         <key>USE_HFS+_COMPRESSION</key>
         <false/>
         <key>VERSION</key>
-        <string>1.0</string>
+        <string>{full-version}</string>
       </dict>
       <key>TYPE</key>
       <integer>0</integer>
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM

PR #270, opened 2020-11-25
Create PR tester for wix installers

gdams:


aahlenst (approved):
Might be helpful if the generated artifacts were collected. So one could test/inspect them easily.

aahlenst:
What about caching the downloaded artifacts? The less to download the better. https://github.com/actions/cache

PR #277, opened 2020-12-16
add aarch64 windows installer support

gdams:
fixes: https://github.com/AdoptOpenJDK/openjdk-installer/issues/276

Test build: https://ci.adoptopenjdk.net/view/work-in-progress/job/create_installer_windows_aarch64/10/

Confirmed to be working on Surface Pro X

### DIFF
--- a/wix/Build.OpenJDK_generic.cmd
+++ b/wix/Build.OpenJDK_generic.cmd
@@ -7,7 +7,7 @@ REM PRODUCT_MAINTENANCE_VERSION=0
 REM PRODUCT_PATCH_VERSION=0
 REM PRODUCT_BUILD_NUMBER=28
 REM MSI_PRODUCT_VERSION=11.0.0.28
-REM ARCH=x64|x86-32 or both "x64 x86-32"
+REM ARCH=x64|x86-32|arm64 or all "x64 x86-32 arm64"
 REM JVM=hotspot|openj9 or both JVM=hotspot openj9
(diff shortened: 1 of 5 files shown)


wix/Build.OpenJDK_generic.cmd
```diff
@@ -26,14 +26,37 @@ IF NOT DEFINED JVM SET ERR=8
...
+                                            )
+                                        )
+                                    )
+                                )
+                            )
+                        )
+                    )
+                )
+            )
+        )
```
aahlenst, line 59: Haven't we reached the point where it would make sense considering using something else, like PowerShell?

gdams: A refactor to PS would be great. I'd prefer to get the arm changes in for now though and work on the PS more slowly.

karianna (approved):
LGTM

wix/Build.OpenJDK_generic.cmd
douph1: I think we can drop and forget all the logic wich assume we build two different things in one run.
I don't see job launched with two arch.
Same for both jvm. One build for hotspot and openj9 at the same time is no more used.
Same for jre and jdk. Not build in same job run since :
>65e6abcc (Morgan Davies      2020-06-05 09:26:03 +0100  475)                 context.println("We have a JRE. Running another installer for it...")

PR #284, opened 2021-01-15
Update supported distributions for Linux packages

aahlenst:
* Debian: Remove Jessie (EOL), add upcoming Bullseye.
* Ubuntu: Remove Eoan (EOL), add upcoming Hirsute.
* Fedora: Remove 31 (EOL), add upcoming 34, aarch64 and armhfp for >= 33.


sxa (dismissed):
Hmmm do we really want to be listing ones that haven't gone to GA yet?

sxa (commented):
Ooops that was supposed to just be a comment rather than an approval ...

karianna (commented):
Same thoughts as sxa here

gdams (approved):

aahlenst:
Well, I've been doing this almost since the beginning. Otherwise we have to re-run and re-upload a lot as soon as they come out.

PR #285, opened 2021-01-18
Separate RPM generation for SUSE from Red Hat

aahlenst:
So far, we were able to generate SUSE and Red Hat RPMs from the same configuration (with some cheating around paths). This is no longer tenable because also the list of dependencies started to diverge. For example, the DejaVu fonts are called dejavu-sans-fonts in Red Hat flavours while SUSE calls them dejavu-fonts.

Because we're in a release week: The change fixes the broken generation of SUSE packages. It does not affect the release builds because Linux package generation isn't run as part of the build (only during nightlies).


karianna (approved):
lgtm

PR #293, opened 2021-02-25
Add workaround to release 8u282-b08 (arm/aarch64)

aahlenst:
The file names on GitHub do not follow the standard conventions.

Will back out the change once the packages have been created.

PR #294, opened 2021-02-26
Make distribution type matching more robust

aahlenst:
Follow-up to #293, but this one can stay because it's a generally applicable improvement.


karianna (approved):
LGTM

PR #295, opened 2021-02-27
Undo workaround to release 8u282-b08 (arm/aarch64)

aahlenst:
Removes temporary changes introduced with #293 that were required because the release filenames did not match expectations.


karianna (approved):
LGTM

PR #298, opened 2021-03-17
Deal with missing U in file names for 16

aahlenst:
Because https://github.com/AdoptOpenJDK/openjdk-build/issues/2524.


sxa (approved):
Seems good although we're should allow it to pick up either version of the name in the future

Issue #301, opened 2021-03-22
Windows installer encoding


douph1:
https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/create_installer_windows/13072/console
on
build-ibmcloud-win2012r2-x64-1
Artifact : OpenJDK8U-jre_x64_windows_hotspot_2021-03-22-18-05.msi

![encoding](https://user-images.githubusercontent.com/45632394/112053955-47eb0e00-8b55-11eb-836d-ded9f276a566.png)

PR #302, opened 2021-03-22
Update Build.OpenJDK_generic.cmd

douph1:
Fix translation lost encoding


douph1:
for #301

aahlenst (approved):

douph1:
I have just push-force same commit with better commit log message

PR #308, opened 2021-04-23
add Name element to Component tree

gdams:
I ran our windows installers through [Accessibility Insights](https://accessibilityinsights.io/en/) and it reported that we should include a Name element for all of the components in the tree.


gdams:
That didn't work, I've asked at https://github.com/wixtoolset/issues/discussions/6435

PR #314, opened 2021-05-18
Openj9 spec addition for linuxNew

jerboaa:
@AdamBrousseau @smlambert FYI.


AdamBrousseau (approved):
fwiw, I've tried this locally and it works for me except the 3 source downloads.
Thanks for this work @jerboaa

smlambert (approved):
thanks @jerboaa !

PR #315, opened 2021-05-24
Delete pkgbuild/notarize directory

gdams:
We no longer need this code because of the eclipse codesigning scripts


karianna (approved):
LGTM

Issue #329, opened 2021-08-05
MSI: RegKey "CurrentVersion" is not set on latest Adoptium JDK8 build


AdamBrousseau:
### To reproduce
- Install 8u292 with install option `JavaSoft (Oracle) Registry Keys`
- Install 8u302 with same install option

### Observed behaviour
- After 292 install, entry for `HKEY_LOCAL_MACHINE/SOFTWARE/JavaSoft/Java Development Kit` `CurrentVersion=1.8`
- After 302 install, this entry is not there. The child entries are there (i.e 1.8 folder and 8.0.302.8 folder)

### Expected behaviour
- CurrentVersion should still be set to 1.8

### Screenshots
![image](https://user-images.githubusercontent.com/6430052/128393890-0cdb14d3-2501-42fb-b25e-b8eb664f69c5.png)

![image](https://user-images.githubusercontent.com/6430052/128393667-b0724b88-5eb3-4eb3-8e22-ae2258afa7bc.png)

### Suspected change
https://github.com/adoptium/installer/commit/e178aefa05d159b0f031ebfff3d543ee8f6103a7
https://github.com/adoptium/installer/pull/307
https://github.com/adoptium/installer/issues/300

I don't quite understand this condition
```
<Condition><![CDATA[JAVASOFT_CURRENTVERSION <> "1.8" AND JAVASOFT_CURRENTVERSION < $(var.ProductMajorVersion)]]></Condition>
```
Says if current version (installed) is 1.8 and current version is less than major version we're building (8), then set CurrentVersion. So it should work? Not sure why it doesn't or why that condition is there. But, this is a bit beyond me.

If I change it to the following, it works
```
<Condition><![CDATA[$(var.ProductMajorVersion) <> "8"]]></Condition>
```
But If I install one with the patch, with no previous install, I don't get the reg values. So I'm not sure what's going on.

AdamBrousseau:
@douph1 Do you have a comment on this issue? We likely need to respin the MSI's again.

douph1:
This condition is here to overide the reg key only if the new major version from the running installer is greater than the previous one installed.
>JAVASOFT_CURRENTVERSION 

is read from the regritry.
If the new var.ProductMajorVersion is greater ( installing java17 over java16) then we overiding the reg key.
If we install java11 after java16 .. the key is unchanged.
JAVASOFT_CURRENTVERSION must pointing to the highest version.
>https://www.oracle.com/java/technologies/javase/runtime-windows.html
[...]

PR #341, opened 2021-08-17
reg key CurrentVersion not writted with jdk8 #329

douph1:
The reg key is now added if not already present.
If present and already "1.8" skip recreating.
If present and <> "1.8" check if major version is newer.

All this tests are depended from the base key wich is different for 1.8 and newer and different for JRE vs JDK
><Property Id="ORACLE_JAVASOFT_BASE_KEY" Value="$(var.OracleJavasoftBaseKey)" />


karianna (approved):
LGTM

PR #356, opened 2021-11-03
Add archive artefact stage to linux new installers

gdams:
This PR archives the generated deb/rpm files in jenkins.

### DIFF
--- a/linuxNew/Jenkinsfile
+++ b/linuxNew/Jenkinsfile
@@ -24,6 +24,11 @@ pipeline {
             }
         }
     }
+    post {
+        always {
+            archiveArtifacts artifacts: '*/*/build/ospackage/*,*/*/build/reports/**'
+        }
+    }
 }
 
 def setup() {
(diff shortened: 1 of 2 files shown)


linuxNew/Jenkinsfile
```diff
@@ -50,7 +51,7 @@ def buildAndTest() {
             break;
     }
     for (os in ["Debian", "RedHat", "Suse"]) {
-        sh("./gradlew clean packageJdk${os} checkJdk${os} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION}")
+        sh("./gradlew packageJdk${os} checkJdk${os} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION}")
```
sophia-guo, line 54: It's been updated. So we don't need the for loop any more. 

`./gradlew --parallel check -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION} --stacktrace` should be good?

sophia-guo:
Will also add check-ca-certificates job in a separate PR.

linuxNew/Jenkinsfile
gdams: In the most recent refactor I've broken the jobs down into sub-jobs that in parallel for each version/distro.

sophia-guo: A second thought, with the parallel for each version is it possible or easy to only generate one version of installers with jenkinsfile?

PR #359, opened 2021-11-09
support s390, ppcle on redhat

sophia-guo:


jerboaa (approved):
LGTM

jerboaa:
@sophia-guo Since there are test failures, I suggest to fix them first before merging. Issues I can think of are:

1. The jobs should first build an SRPM from the spec `rpmbuild --bs <spec.file>`. Then as a second step use the SRPMs to build the binary. `rpmbuild --rebuild <SRPM>`
2. While the building of the ppc64le/s390x bits can happen on `x86_64` using `--target`, the actual tests need to run on the respective `s390x` and `ppc64le` machines. Both will need to have `docker` installed there (`podman` support could be considered too)
3. JDK 8 doesn't have s390x builds. For Hotspot only Zero Interpreter is supported on that architecture which is slow. I'd suggest to disabled for JDK 8. JDK 11 and 17 work fine.
4. JDK 16 isn't any longer, I'd drop it in CI favor of JDK 17. Not sure why the ppc64le build failed, but I supposed if step 1 was to be done, it would already fail on the SRPM build.

The first point has the advantage that users can rebuild the srpm on their local system if they wanted to (with a spec file fix). Plus, it automatically validates that the SRPM is working.

sophia-guo:
Thanks @jerboaa! Yes, I have converted the PR as draft as there are still issues need to be address.

> 1. The jobs should first build an SRPM from the spec rpmbuild --bs <spec.file>. Then as a second step use the SRPMs to build the binary. rpmbuild --rebuild <SRPM>.
The first point has the advantage that users can rebuild the srpm on their local system if they wanted to (with a spec file fix). Plus, it automatically validates that the SRPM is working.

Does this mean we will also need to archive SRPM?

> While the building of the ppc64le/s390x bits can happen on x86_64 using --target, the actual tests need to run on the respective s390x and ppc64le machines. Both will need to have docker installed there (podman support could be considered too)

We will double check if [Testcontainers](https://www.testcontainers.org/) supports s390x and ppc64le or not.

> JDK 8 doesn't have s390x builds. For Hotspot only Zero Interpreter is supported on that architecture which is slow. I'd suggest to disabled for JDK 8. JDK 11 and 17 work fine.

Yes, jdk8 s390x will be removed.

karianna:
@sophia-guo & @jerboaa - Thanks so much for taking this over after I dropped it.  I'm out most of this week and maybe part next week but happy to help test etc however I can when I'm back.  Once again thanks!

sophia-guo:
As different version support different architecture will update Dockerfile to  produce rpm for different architectures on x86_64 per version

sxa (approved):

sophia-guo:
@jerboaa Updated with 

- Remove jdk16, which is replaced by jdk17 
- Build SRPM and copy to destination folder
- Remove s390 for jdk8.

Issue #358, opened 2021-11-08
Installer naming for different distro, i.e. SUSE-based systems and RedHat


sophia-guo:
The linuxNew installer generates the installer with same name for RedHat and SUSE. Take jdk11 for example: 
temurin-11-jdk-11.0.13.0.0.8-1.x86_64.rpm under different folders. If we'd like to release them to github repo probably we need to name them add the distro information.

Any suggestions?

sxa:
It's reasonably common for the names to include the distribution. For example on one of my RHEL7 machines the gzip package is called `gzip-1.5-10.el7.x86_64` where the `el7` is for RHEL7. We should follow a similar convention even if we don't use the same shortname (since we're not going to ship separate RPMs for each version of RHEL. I think my preference would be e.g. `temurin-11-jdk-11.0.13.0.0.8-1.rhel.x86_64.rpm` and `temurin-11-jdk-11.0.13.0.0.8-1.sles.x86_64.rpm` or just `.rh.` and `.sl.` in there. If we did want to explicitly have a version in there then we could go for the lowest we would suppot e.g. `.el6`

sophia-guo:
Close it as we are pushing directly to artifactory, which can handle this staying in different directory.

PR #360, opened 2021-11-11
update naming

sophia-guo:
Related #358 


gdams:
This technically is no longer needed as we are pushing directly to artifactory now? Shall we close?

sophia-guo:
Yes, close it.

Issue #363, opened 2021-11-12
Identify suitable cacerts package for .deb packages


sxa:
The initilal issue #330 was prototyped using a separate `adoptium-ca-certificates` package but based on [a comment](https://adoptium.slack.com/archives/C9C272Z1T/p1636727818021100?thread_ts=1636726678.017800&cid=C9C272Z1T) in slack we may be able to make `temurin-17` etc. provide the virtual package `java17-runtime-headless` in order to unblock the cyclic dependency described [in that issue](https://github.com/adoptium/installer/issues/330#issuecomment-896308438) rather than having to provide our own separate cacerts package. This needs to be verified, but would be a cleaner solution than the original proposal if it works.

PR #370, opened 2021-11-15
Change dependencies of Debian package to use ca-certificates-java instead of adoptium version

sxa:
Fixes [#deb](https://github.com/adoptium/installer/issues/363), but this currently causes the tests to fail as it MUST be installed with `apt-get install ./temurin*deb` instead of just `dpkg` unless you have another java on the machine, so I'm holding this in draft for now to see if we can make the tests work.

### DIFF
--- a/linuxNew/Jenkinsfile
+++ b/linuxNew/Jenkinsfile
@@ -59,8 +59,8 @@ def buildAndTest() {
         default: //Adoptium API 
             break;
     }
-    sh("./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION}")
-    archiveArtifacts artifacts: '**/*/build/ospackage/*,**/*/build/reports/**'
+    sh("./gradlew :ca-certificates:package packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION}")
+    archiveArtifacts artifacts: '**/*/build/ospackage/*,**/*/build/reports/**,**/build/**/adoptium-ca-certificates*.deb'
 }
 
 def uploadArtifacts() {
--- a/linuxNew/jdk/debian/build.gradle
+++ b/linuxNew/jdk/debian/build.gradle
@@ -24,7 +24,7 @@ version "1.0.0-SNAPSHOT"
  *
  * **Attention**: If you alter the list, check if the class DebianFlavours needs to be updated, too.
  */
-def deb_versions = ["bullseye", "bionic", "focal", "groovy", "hirsute"]
+def deb_versions = ["stretch", "buster", "bullseye", "bionic", "focal", "groovy", "hirsute"]
 
 java {
 	sourceCompatibility = JavaVersion.VERSION_1_8
--- a/linuxNew/jdk/debian/build.gradle
+++ b/linuxNew/jdk/debian/build.gradle
@@ -24,7 +24,7 @@ version "1.0.0-SNAPSHOT"
  *
  * **Attention**: If you alter the list, check if the class DebianFlavours needs to be updated, too.
  */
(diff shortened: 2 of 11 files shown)


nudgegoonies:
Why do you want this? ca-certificates-java cannot work:
```
head: cannot open '/etc/ssl/certs/java/cacerts' for reading: No such file or directory
/var/lib/dpkg/info/ca-certificates-java.postinst: line 101: java: command not found
dpkg: error processing package ca-certificates-java (--configure):
 installed ca-certificates-java package post-installation script subprocess returned error exit status 127
dpkg: dependency problems prevent configuration of temurin-8-jdk:
 temurin-8-jdk depends on ca-certificates-java; however:
  Package ca-certificates-java is not configured yet.

dpkg: error processing package temurin-8-jdk (--configure):
 dependency problems - leaving unconfigured
Processing triggers for libc-bin (2.31-13+deb11u2) ...
Processing triggers for ca-certificates (20210119) ...
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...

/etc/ca-certificates/update.d/jks-keystore: 82: java: not found
E: /etc/ca-certificates/update.d/jks-keystore exited with code 1.
done.
Errors were encountered while processing:
 ca-certificates-java
 temurin-8-jdk
E: Sub-process /usr/bin/dpkg returned an error code (1)
```
Please stay on adoptium-ca-certificates as this is the solution to the ca-certificates-java problem.

sxa:
Thanks for the comments. Which distribution were you seeing that failure on? I believe that of the currently supported distributions this is only an issue for Debian Buster which is still in support for just under a year. So we have the option to revisit that decision (the slack conversation where it was discussed last week is [here](https://adoptium.slack.com/archives/C9C272Z1T/p1636726678017800)) and list some alternative workaround for Buster and this PR was to evaluate that and get feedback.

nudgegoonies:
I used debian:bullseye docker image and installed with "apt install ./FILENAME" from the jenkins build.

As far as i know using ca-certificates-java will only work, if ca-certificates java supports "foreign" JDKs/JREs.
This was my old MR for AdoptOpenJDK in the past but it was never merged:
https://salsa.debian.org/nudgegoonies-guest/ca-certificates-java/-/commit/9a48ee3879919cd3583756313c5fd5a76f1388a7

I updated it to support Temurin but i expect that debian does not want to support foreign JDKs except oracle which is already included:
https://salsa.debian.org/nudgegoonies-guest/ca-certificates-java/-/commit/9a48ee3879919cd3583756313c5fd5a76f1388a7

The problem is still, that a JDK must be at least unpacked (debian package state) within one of the /usr/lib/jvm paths that ca-certificates-java can use one of these JDKs to run.
Temporary installing debians OpenJDK or Oracle JDK to allow installation of Temurin is not a solution!

nudgegoonies:
The whole trick seems to be, that this postinst script (that means that temurin and ca-certificates-java are already unpacked) can find a working java executable here because the script only looks for debians openjdk and oracle jdks:
https://salsa.debian.org/java-team/ca-certificates-java/-/blob/master/debian/ca-certificates-java.postinst#L21 (current master)
https://salsa.debian.org/java-team/ca-certificates-java/-/blob/debian/20190909/debian/postinst#L26 (bullseye)

The error message from my output above comes from this function:
https://salsa.debian.org/java-team/ca-certificates-java/-/blob/debian/20190909/debian/postinst#L87
And here is the first use of java which fails because the setup_path function hasn't added a temurin jdk to the path and then the postinst script fails:
https://salsa.debian.org/java-team/ca-certificates-java/-/blob/debian/20190909/debian/postinst#L101

nudgegoonies:
With my MR it would work. I tested it in fresh docker-containers. But that fix must also be backported to the stretch, buster and bullseye versions of ca-certificates-java.

sxa:
Yeah I've had a rethink on this and I'm going to give up on this for now. We can revisit later if it makes sense to look at it further. Thanks for your input @nudgegoonies. We've got the package being correctly built now and I've just tested it - just need to expose it.

PR #373, opened 2021-11-16
add GPG key to rpm build

gdams:
This PR add's the ability to pass a private GPG key to the docker environment to allow RPM signing to take place

Tested in https://ci.adoptopenjdk.net/job/gdams-packages-linux-pipeline/16/

### DIFF
--- a/linuxNew/Jenkinsfile
+++ b/linuxNew/Jenkinsfile
@@ -38,6 +38,8 @@ pipeline {
 
 def setup() {
     cleanWs()
+    // Docker --mount option requires BuildKit
+    env.DOCKER_BUILDKIT=1
     env._JAVA_OPTIONS="-Xmx4g"
     checkout scm
 }
@@ -54,9 +56,10 @@ def buildAndTest() {
     // Install Adoptium GPG key for RPM signing
     withCredentials([file(credentialsId: 'adoptium-artifactory-gpg-key', variable: 'GPG_KEY')]) {
         if (DISTRO != "Debian") {
-            SIGN_CMD="-PGPG_KEY=${GPG_KEY}"
+            sh("./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION} -PGPG_KEY=${GPG_KEY} || true")
+        } else {
+            sh("./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION} || true")
         }
-        sh("./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION} ${SIGN_CMD} || true")
     }
     archiveArtifacts artifacts: '**/build/ospackage/*,**/build/reports/**,**/packageTest/dependencies/deb/*'
 }
--- a/linuxNew/Jenkinsfile
+++ b/linuxNew/Jenkinsfile
@@ -4,14 +4,14 @@ pipeline {
     agent {
         label NODE_LABEL
     }
-    tools {
-        jdk "JDK11"
-    }
     stages {
         stage('Linux Installers') {
             matrix {
                 agent {
                     label NODE_LABEL
+                }
+                tools {
+                    jdk "JDK11"
                 }
                 axes {
                     axis {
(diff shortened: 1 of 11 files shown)


linuxNew/Jenkinsfile
```diff
@@ -59,6 +60,14 @@ def buildAndTest() {
     archiveArtifacts artifacts: '**/*/build/ospackage/*,**/*/build/reports/**,*,**/ospackage/adoptium-ca-certificates*.deb'
 }
 
+def setupGPG() {
```
karianna, line 63: Should this be made optional for others somehow?

linuxNew/jdk/redhat/src/main/packaging/entrypoint.sh
```diff
@@ -5,6 +5,8 @@ set -euox pipefail
 # into it without changing its ownership (which could render the folder or its contents inaccessible to the host user),
 # add the user builder to the group with the same GID as the host user.
 HOST_USER_GID=$(stat -c "%g" /home/builder/out)
+# gpg: signing failed: Inappropriate ioctl for device
```
sxa, line 8: Is this intended to indicate that this error occurs if the next line is not in place? Would be good to make that clear in the comment.

linuxNew/jdk/redhat/src/main/packaging/temurin/8/temurin-8-jdk.spec
```diff
@@ -242,5 +242,5 @@ fi
 /usr/lib/tmpfiles.d/%{name}.conf
 
 %changelog
-* Fri Aug 31 2021 Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org> 8.0.302.0.0.8-1.adopt0
```
sxa, line 245: Nice spot :-)

linuxNew/jdk/suse/src/main/packaging/entrypoint.sh
```diff
@@ -5,6 +5,8 @@ set -euox pipefail
 # into it without changing its ownership (which could render the folder or its contents inaccessible to the host user),
 # add the user builder to the group with the same GID as the host user.
 HOST_USER_GID=$(stat -c "%g" /home/builder/out)
+# gpg: signing failed: Inappropriate ioctl for device
```
sxa, line 8: As per earlier comment on the redhat file

linuxNew/jdk/suse/build.gradle
```diff
@@ -67,14 +68,26 @@ task packageJdkSuse {
...
+		if (gpgKey) {
+			project.exec {
+				workingDir "src/main/packaging"
+				commandLine "docker", "build",
+					"-t", "adoptium-packages-linux-jdk-suse",
+					"--secret", "id=gpg,src=${gpgKey}",
+					"-f", "Dockerfile",
+					getProjectDir().absolutePath + "/src/main/packaging"
+			}
+		} else {
```
sxa, line 80: Under what circumstances do we want this to succeed if no gpg key is available? At the very least I think I'd want an option to say "Abort if none is available" to avoid accidentally producing and distributing unsigned ones.

sxa (changes requested):
Small number of questions points, but in general this looks good.

linuxNew/jdk/suse/build.gradle
gdams: I've left this else statement in so that people can use other methods to sign the generated RPMs. It's also used in GitHub actions because we don't want our GPG key sitting on those executors.

linuxNew/jdk/redhat/src/main/packaging/entrypoint.sh
gdams: updated

linuxNew/jdk/suse/build.gradle
sxa: Agreed - I wouldn't be in favour of removing it but having an OPTION to ensure our own pipelines will abort if it isn't in place would seem a good idea.

sxa: (Which is along the same lines as what @karianna said in [his comment](https://github.com/adoptium/installer/pull/373#discussion_r750526430) :-) ). I'm otherwise ok with it

sxa (approved):
I'll approve, but if this is going to me merged without a suitable trap in the case of an absence of GPG signing I want a separate issue to cover that with an intention of resolving it.

gdams:
> I'll approve, but if this is going to me merged without a suitable trap in the case of an absence of GPG signing I want a separate issue to cover that with an intention of resolving it.

We could certainly add a test for this in a follow-up PR

PR #374, opened 2021-11-16
Do not fail jenkins if test fails so rpm can be archived

### LINKED ISSUE #362: Make sure archive rpm generated even if the testcontainer tests failed.
Current testcontainer tests don't work with ppc64le and s390x architecture we'd like to update to make sure those ppc64le and s390x rpm will be archived and released even if the tests fail.

sophia-guo:
Fix #362 

### DIFF
--- a/linuxNew/Jenkinsfile
+++ b/linuxNew/Jenkinsfile
@@ -55,7 +55,7 @@ def buildAndTest() {
         default: //Adoptium API 
             break;
     }
-    sh("./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION}")
+    sh("./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION} || true")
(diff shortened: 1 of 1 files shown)


sophia-guo:
Originally works no rpm be archived due to the test failures. https://ci.adoptopenjdk.net/view/work-in-progress/job/adoptium-packages-linux-pipeline/51/

Now: https://ci.adoptopenjdk.net/view/work-in-progress/job/adoptium-packages-linux-pipeline/53/

smlambert (approved):
LGTM

gdams:
I'm not sure about this change, can we not skip the tests for other platforms (like we do with aarch64 in the setup stage)

smlambert:
I'd advised that we do something temporarily while we work out the testing for other platforms (an issue for fixing https://github.com/adoptium/installer/issues/375).  Skipping tests if its easy/doable is likely a better approach  as a temporary state.

sxa:
I think there's also an issue with the tests being somewhat flakey. I've seen that in some of the runs I'm doing. One time they'll work, the next time they won't. When I've looked at them they've generally been the ones for the Red Hat rpms, but not necessarily the same one on each run, so this bypasses that problem for now, but we do need to make sure later that we're not "hiding" any real problems with this.

PR #377, opened 2021-11-17
add aarch64 support and make github actions upload test results

gdams:


sophia-guo:
No sure if `aarch64` works may try `arm64`?

PR #388, opened 2021-11-23
Upload RPM to artifactory

sophia-guo:


smlambert (approved):
thanks @sophia-guo, looks good !  (and as discussed, build-info, parameterization for publish=true and further refinements can come as follow-up PRs).

sophia-guo:
Following PR will do the spec with separate json files.

PR #406, opened 2021-12-16
Only enable non-x64 packaging stage to debian distr

sophia-guo:


gdams:
I'm not convinced this would work? Say I kicked off a build for Rhel based only wouldn't it still run the Debian steps?

sophia-guo:
If distro is redhat, then no debian steps will run.

https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/66/

gdams (approved):

PR #410, opened 2021-12-30
RPM - add missing RHEL and CentOS distros

gdams:
Lots of people have requested CentOS 7 and RHEL packages to be uploaded


sxa (approved):
Assuming this is all that's required, I definitely approve of this change (as mentioned in https://github.com/adoptium/blog.adoptium.net/pull/115) since CentOS8 is EOL and 7 is supported until June 2024.

PR #414, opened 2021-12-31
DEB: add libatomic1 dependency for armhf

gdams:
See https://github.com/AdoptOpenJDK/openjdk-docker/issues/586 for context

fixes:

```bash
root@2bca16a673f9:/# java -version
/usr/bin/java: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
```

Already defined for the RPMs in https://github.com/adoptium/installer/blob/master/linuxNew/jdk/redhat/src/main/packaging/temurin/8/temurin-8-jdk.spec#L79-L81


karianna (approved):
LGTM

PR #423, opened 2022-02-01
Update packaging for jdk-11.0.14+9

sxa:
All architectures are now available for JDK11.0.14+9.

Signed-off-by: Stewart X Addison <sxa@redhat.com>

PR #425, opened 2022-02-01
Update packaging for jdk-17.0.2+8

sxa:
Draft as arm32 and s390x are not yet published (Should be later today)

Also note that unlike #423 I have not included the old version in the Debian changelog for now as it is failing the checks. We'll need to do something consistent between 8 and 11 ...

### DIFF
--- a/linuxNew/jdk/debian/src/main/packaging/temurin/17/debian/changelog
+++ b/linuxNew/jdk/debian/src/main/packaging/temurin/17/debian/changelog
@@ -3,3 +3,9 @@ temurin-17-jdk (17.0.2.0.0+8-1) UNRELEASED; urgency=medium
   * Eclipse Temurin 17.0.2.0.0+8-1 release.
 
  -- Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>  Thu, 01 Feb 2022 14:16:30 +0000
+
+temurin-17-jdk (17.0.1.0.0+12-1) UNRELEASED; urgency=medium
+
+  * Eclipse Temurin 17.0.1.0.0+12-1 release.
+
+ -- Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>  Thu, 23 sep 2021 15:32:48 +0000
--- a/linuxNew/jdk/debian/src/main/packaging/temurin/17/debian/changelog
+++ b/linuxNew/jdk/debian/src/main/packaging/temurin/17/debian/changelog
@@ -1,5 +1,5 @@
-temurin-17-jdk (17.0.1.0.0+12-1) UNRELEASED; urgency=medium
+temurin-17-jdk (17.0.2.0.0+8-1) UNRELEASED; urgency=medium
 
-  * Eclipse Temurin 17.0.1.0.0+12-1 release.
+  * Eclipse Temurin 17.0.2.0.0+8-1 release.
 
- -- Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>  Thu, 23 sep 2021 15:32:48 +0000
+ -- Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>  Thu, 01 Feb 2022 14:16:30 +0000
--- a/linuxNew/jdk/debian/src/main/packaging/temurin/17/debian/rules
+++ b/linuxNew/jdk/debian/src/main/packaging/temurin/17/debian/rules
@@ -3,16 +3,16 @@
 pkg_name = temurin-17-jdk
 priority = 1161
 jvm_tools = jaotc jar jarsigner java javac javadoc javap jcmd jconsole jdb jdeprscan jdeps jfr jhsdb jimage jinfo jlink jmap jmod jpackage jps jrunscript jshell jstack jstat jstatd keytool rmid rmiregistry serialver jexec jspawnhelper
-amd64_tarball_url = https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.1%2B12/OpenJDK17U-jdk_x64_linux_hotspot_17.0.1_12.tar.gz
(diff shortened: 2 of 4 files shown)


karianna (approved):
LGTM

sxa:
Switching to draft as per initial comment until s390x and arm32 are released.

sxa:
Removing from draft again now that the two extra Linux platforms for 17 have shipped.

sxa:
Not immediately obvious why these are failing - requires input from some of the requsted reviewers to understand what's going on. Looks like it's objecting to some missing checksum files but not sure if that's the underlying cause or why it has happened.

sxa:
> > Not immediately obvious why these are failing - requires input from some of the requsted reviewers to understand what's going on. Looks like it's objecting to some missing checksum files but not sure if that's the underlying cause or why it has happened.
> 
> The run Martijn linked is a 404 not found on the download which leads to file missing. I'd try to re-run the action to see if it can catch them this time. Could be a network hiccup on that machine, or they weren't available at the time it ran.

As per the issue description that would have been kicked off before the arm32 and s390x builds had been released. The subsequent re-run (hoipefully) didn't have that same error. There will be another one running now since I've pushed up an update to th eDebian changelog in this PR.

> Looks like there was a second run where the expected tarball isn't found in the sha256 check process. Maybe it's expecting the wrong name for that particular version-arch combo? I recall before there was a build + sha256 combo where the sha file was referencing the wrong tarball name. Seems like it could have occurred again.

I think that was specific to JDK8/aarch32 becuase of issues caused by the alternate repository which shouldn't be applicable here.

jiekang:
> > > Not immediately obvious why these are failing - requires input from some of the requsted reviewers to understand what's going on. Looks like it's objecting to some missing checksum files but not sure if that's the underlying cause or why it has happened.
> > 
> > 
> > The run Martijn linked is a 404 not found on the download which leads to file missing. I'd try to re-run the action to see if it can catch them this time. Could be a network hiccup on that machine, or they weren't available at the time it ran.
> 
> As per the issue description that would have been kicked off before the arm32 and s390x builds had been released. The subsequent re-run (hoipefully) didn't have that same error. There will be another one running now since I've pushed up an update to th eDebian changelog in this PR.
> 
> > Looks like there was a second run where the expected tarball isn't found in the sha256 check process. Maybe it's expecting the wrong name for that particular version-arch combo? I recall before there was a build + sha256 combo where the sha file was referencing the wrong tarball name. Seems like it could have occurred again.
> 
> I think that was specific to JDK8/aarch32 becuase of issues caused by the alternate repository which shouldn't be applicable here.

Looks like a similar but unrelated issue. The sha file has:
```
302caf29f73481b2b914ba2b89705036010c65eb9bc8d7712b27d6e9bedf6200  target/linux/aarch64/hotspot/OpenJDK17U-jdk_aarch64_linux_hotspot_17.0.2_8.tar.gz
···

I imagine it should be like s390x one without the path:
···
383ac8ad392036bedab9a08eb55395b95593a6cc268c422a2bab53f0977a4c54  OpenJDK17U-jdk_s390x_linux_hotspot_17.0.2_8.tar.gz
···

**Comment by sxa** [comment]
> > > > Not immediately obvious why these are failing - requires input from some of the requsted reviewers to understand what's going on. Looks like it's objecting to some missing checksum files but not sure if that's the underlying cause or why it has happened.
> > > 
> > > 
> > > The run Martijn linked is a 404 not found on the download which leads to file missing. I'd try to re-run the action to see if it can catch them this time. Could be a network hiccup on that machine, or they weren't available at the time it ran.
> > 
> > 
> > As per the issue description that would have been kicked off before the arm32 and s390x builds had been released. The subsequent re-run (hoipefully) didn't have that same error. There will be another one running now since I've pushed up an update to th eDebian changelog in this PR.
> > > Looks like there was a second run where the expected tarball isn't found in the sha256 check process. Maybe it's expecting the wrong name for that particular version-arch combo? I recall before there was a build + sha256 combo where the sha file was referencing the wrong tarball name. Seems like it could have occurred again.
> > 
> > 
> > I think that was specific to JDK8/aarch32 becuase of issues caused by the alternate repository which shouldn't be applicable here.
> 
> Looks like a similar but unrelated issue. The sha file has:
> 
> ```
> 302caf29f73481b2b914ba2b89705036010c65eb9bc8d7712b27d6e9bedf6200  target/linux/aarch64/hotspot/OpenJDK17U-jdk_aarch64_linux_hotspot_17.0.2_8.tar.gz
> ```
> 
> I imagine it should be like s390x one without the path:
> 
> ```
> 383ac8ad392036bedab9a08eb55395b95593a6cc268c422a2bab53f0977a4c54  OpenJDK17U-jdk_s390x_linux_hotspot_17.0.2_8.tar.gz
> ```

That's odd - unclear what's happened to make that only affect some of them, but I've manually updated the one causing the problem and will re-run the checks.

**Comment by jiekang** [comment]
Now looks like the temurin-8/11/17-redhat builds all fail on the same test:

```
DnfOperationsTest > centos:8 FAILED
    org.opentest4j.AssertionFailedError at DnfOperationsTest.java:54
```

This is an attempt to update the system with `dnf update -y`.
(https://github.com/adoptium/installer/blob/163e7530fcd500e908f8c56a5fd8ba5c4662d7aa/linuxNew/jdk/redhat/src/packageTest/java/packaging/DnfOperationsTest.java)

Would need better log output to know why, or maybe someone has past experience with this in a GH actions run.

gdams (approved):

karianna:
@gdams 3 checks failed - do we still need to fix something here?

sophia-guo:
%changelog part of both redhat and suse also need to update.

sophia-guo:
NM. I saw the log has been updated in https://github.com/adoptium/installer/pull/423

PR #428, opened 2022-02-03
Update packaging for jdk8u322-b06

gdams:
draft until arm32 is released


theofficialgman:
just wanted to note this here because its been an issue for some projects which use java which introduces binary incompatibility with multiple projects. the problematic change is a change to `ManifestEntryVerifier` where a parameter is now required, while it was not before these commits

commits:
https://github.com/openjdk/jdk8u/commit/9edb3c788c590ca30b1bc94539035cf1e7c3c119
https://github.com/openjdk/jdk11u/commit/9512c471e9ab0830d8d5b2275803f59a0f073fcb
https://github.com/openjdk/jdk17u/commit/f89c59f9d151eea3f747155b129e17c9e6e8ec32

as you can see from the commits above, this change is included in 322+ on jdk8

an example of a project which older versions were all broken by this change was the McModLauncher project (related issue here: https://github.com/McModLauncher/modlauncher/issues/91) which is used by the Forge minecraft mod project among other projects.

PR #437, opened 2022-02-15
mac: Trap case where nether hotspot or openj9 are in the filename

sxa:
Self-explanatory - traps the default condition which showed up before https://github.com/adoptium/ci-jenkins-pipelines/pull/241 went in. Error was an attempt to use an undefined variable.

```
pkgbuild/create-installer-mac.sh: line 79: JVM: unbound variable
```


karianna (approved):
LGTM

PR #438, opened 2022-02-17
create_installer_windows failing with new "temurin" variant support

sxa:
Fix for the following error in [create_installer_windows](https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/create_installer_windows):
```
JVM "temurin" not supported : valid values : hotspot, openj9, dragonwell, hotspot openj9, openj9 hotspot
```
This just maps `JVM=temurin` to `JVM=hotspot` to ensure all the values are the same as before. Happy to entertain a discussion on whether this should be changed in the future but for now this should allow the installer to be generated the same as it was before the changes.

### DIFF
--- a/wix/Build.OpenJDK_generic.cmd
+++ b/wix/Build.OpenJDK_generic.cmd
@@ -70,6 +70,10 @@ IF NOT "%ARCH%" == "x64" (
 	)
 )
 
+REM Update to handle the change of build variant until implications
+REM of setting this to Temurin can be evaluated
+IF "%JVM%" == "temurin" SET JVM=hotspot
+
 IF NOT "%JVM%" == "hotspot" (
 	IF NOT "%JVM%" == "openj9" (
 	    IF NOT "%JVM%" == "dragonwell" (
--- a/pkgbuild/create-installer-mac.sh
+++ b/pkgbuild/create-installer-mac.sh
@@ -50,6 +50,10 @@ do tar -xf "$f";
     *openj9*)
       export JVM="openj9"
     ;;
+    *)
+      echo Cannot identify variant from filename "${f}" - only hotspot or openj9 are allowable
+      exit 1
+    ;;
   esac
 
   # Detect if JRE or JDK


sxa:
@gdams Any idea what's going on with the checks here? Is http://s.sudre.free.fr/ a legitimate source (if nothing else it seems a bit wrong that it's not https)
```
Run brew install --cask packages
  brew install --cask packages
  shell: /bin/bash -e {0}
==> Downloading http://s.sudre.free.fr/Software/files/Packages.dmg
Warning: No checksum defined for cask 'packages', skipping verification.
Warning: macOS's Gatekeeper has been disabled for this Cask
==> Installing Cask packages
==> Purging files for version 1.2.10 of Cask packages
Error: Failure while executing; `hdiutil convert -quiet -format UDTO -o /private/tmp/d20220217-812-3r9z2s/02a5de38825fc3ddbe60b7fa49ca849cab656a74b8b0fb64158ea39f1a2bcff9--Packages.cdr /Users/runner/Library/Caches/Homebrew/downloads/02a5de38825fc3ddbe60b7fa49ca849cab656a74b8b0fb64158ea39f1a2bcff9--Packages.dmg` exited with 1.
Error: Process completed with exit code 1.
```

gdams:
@sxa I'm not entirely sure what's broken, the cask was updated 10 days ago (https://github.com/Homebrew/homebrew-cask/blob/master/Casks/packages.rb) so I wonder if the link is broken, maybe try running it again?

sxa:
> @sxa I'm not entirely sure what's broken, the cask was updated 10 days ago (https://github.com/Homebrew/homebrew-cask/blob/master/Casks/packages.rb) so I wonder if the link is broken, I'll try running it again

Hmm seems ok this time. Hopefully x64 will go through ok too ...

karianna (approved):
lgtm

wix/Build.OpenJDK_generic.cmd
```diff
@@ -70,6 +70,10 @@ IF NOT "%ARCH%" == "x64" (
 	)
 )
 
+REM Update to handle the change of build variant until implications
+REM of setting this to Temurin can be evaluated
+IF "%JVM%" == "temurin" SET JVM=hotspot
```
douph1, line 75: @sxa, @karianna  Why this special case is not handled on the caller side ?
temurin is not a JVM so the JVM arg must not accept "temurin" and hide some other value behind ..

If you set upsteam a vendor name temurin , his place is in VENDOR_BRANDING from this script, not in JVM arg 

I think in caller job you must do somthing like that: 
if (VARIANT == "temurin" )
 VENDOR_BRANDING"=Temurin"
 JVM="hotspot" // default JVM for VARIANT "temurin"
 call createInstallerJob
fi

sxa: The intention of the initial set of changes was to do the minimum required in the pipelines to enable the use of a temurin variant. There will likely be subsequent changes to split this out using other variables but initially this was the simplest option to prove that we could split out hotspot and temurin. After implementing it, we hit this issue and this was the simplest way to fix it which wouldn't involve any risk to changing the contents of the package.

I agree this could be switched to being done in the calling pipelines, and so your proposed change can be covered in the other issue which Martijn has created :-)

Issue #459, opened 2022-05-02
Update linuxNew pipeline to be able to push subset of platforms


smlambert:
Currently the recently added linuxNew pipeline only supports pushing all platforms in one shot.  It should be enhanced to be able to support pushing singular platforms or a subset of them as they become available which better reflects the release process to date.

zdtsw:
To be more clear about the requirement in this ticket:

change from the current input of job

- jdk version
- distro: redhat|debian|suse

to new input:

- jdk version
- arch selection: x64, ppc64le, armv7l, aarch64,  s390x(when jdk!=8) and x32/i386(when distro==debian)

karianna:
https://github.com/adoptium/installer.git is where it's getting the scripts from which is correct.

zdtsw:
i randomly clicked to  https://ci.adoptopenjdk.net/job/adoptium-packages-linux-pipeline/103/
and also https://ci.adoptopenjdk.net/job/adoptium-packages-linux-pipeline/115/

karianna:
Try https://ci.adoptopenjdk.net/job/adoptium-packages-linux-pipeline/119/ and the currently running one https://ci.adoptopenjdk.net/job/adoptium-packages-linux-pipeline/120/ I think they're OK.

jiekang:
Is https://github.com/adoptium/installer/issues/446 addressing the same issue as this one? Maybe close as duplicate?

zdtsw:
> Is #446 addressing the same issue as this one? Maybe close as duplicate?

looks like duplicated issue

smlambert:
Yes both issues appear to be the same, it was raised in release retrospectives and I was unaware that Adam had raised an issue already.  I'll close the earlier one as this one has had more activity on it.

smlambert:
Has this issue been addressed?  Moving it to August plan, but please link relevant PRs if its partially or fully fixed.

zdtsw:
The original PR is https://github.com/adoptium/installer/pull/472
and the job running with PR's code https://ci.adoptopenjdk.net/job/Sophia-adoptium-packages-linux-pipeline/ (up to build number 139)

sxa:
> Has this issue been addressed?

[...]

PR #472, opened 2022-05-19
feat: add capability to run installer per jdk version or distro selection 

zdtsw:
- add  support ARCH with selection but not include "all" 
            - remove support VERSION with value "all"
            - make DISTRO to only "all" "Debian" "RedHat" and "Suse"
            - pass in jdkversion as variable for rpm
            -  default use node pool "dockerBuild&&linux&&x64", debian non x64 use pool "docker&&linux&&${ARCH}"
            - change from matrix to three different stages per each distro
            - enable DOCKER_BUILDKIT on all even for Debian+s390x
            - only do git clone "installer.git" once and stash folder "linux" into different distro build stage
            - remove rpm for Centos8 publish to jfrog (already have rocky8)
            - add support in jenkins for 
               -  enableDebug for more build info
               -  split UPLOAD into "uploadPackage" and "uploadSRCRPM"
               -  can use the same job for test repo or branch (for debug purpose)
               -  add simple description of arch and distro per each job history
               -  set timeout to 2hr
               - 
Screenshot from 2022-06-28 09-04-46

           - add more test coverage:
              - missing test on oraclelinux8, ubi8
              - previous in one batch of bulding all RedHat/Suse RPMs, only test on x86_64
              - now at least one testcase per each arch, if rpm exists after build and keep "yum/dnf/zypper install" only on x86_64(same as before) due to docker host is for x64 only
Test run logs:
RedHat arm32:
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/125/console

RedHat x64:
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/129/console

Suse x86:
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/127/console

Suse aarch64
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/128/console

Suse all
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/136/console

RedHat all
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/139/console

Debian x64
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/138/console

Debian arm64
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/140/console

Ref: https://github.com/adoptium/installer/issues/459

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -1,126 +1,134 @@
-NODE_LABEL = "dockerBuild&&linux&&x64" // Dynamic docker machines from Azure
+/*
+--------------------------------------------
+buildArch   RPM         DEB         Comments
+--------------------------------------------
+x86_64      x86_64      amd64
+armv7l      armv7hl     armhf       arm32
+aarch64     aarch64     arm64       arm64
+ppc64le     ppc64le     ppc64le
+source      src         -           only for SRPM and no need specify as option or target
+s390x       s390x       s390x       only for jdk8+
+*/
+
+env.NODE_LABEL_RPM = "dockerBuild&&linux&&x64" // RedHat + Suse + Debian x64
+env.NODE_LABEL_DEB = "docker&&linux&&${ARCH}"  // Debian non-x64 
+env.PRODUCT = "temurin"
 
 pipeline {
     agent {
-        label NODE_LABEL
+        label NODE_LABEL_RPM
+    }
+    options {
+        timeout(time: 2, unit: 'HOURS') 
     }
     parameters {
-        string(name: 'PRODUCT', defaultValue: 'temurin')
-        string(name: 'JDK_SOURCE', defaultValue: 'Adoptium')
-        choice(name: 'VERSION_FILTER', choices: ['8', '11', '17', '18', 'all'], description: 'Build for specific VERSION')
-        choice(name: 'DISTRO_FILTER', choices: ['all', 'Debian', 'RedHat', 'Suse'], description: 'Build for specific Distro')
-        booleanParam(name: 'UPLOAD', defaultValue: false, description: 'Tick this box to upload the deb/rpm files to Artifactory')
+        choice(name: 'VERSION', choices: ['8', '11', '17', '18'], description: 'Build for specific JDK VERSION')
+        choice(name: 'ARCH', choices: ['x86_64', 'armv7l', 'aarch64', 'ppc64le', 's390x', 'all'], description: "Build for specific platform\n s300x not for VERSION 8\n")
+        choice(name: 'DISTRO', choices: ['all', 'Debian', 'RedHat', 'Suse'], description: 'Build for specific Distro\n Select RPM builds for RedHat and Suse')
+        booleanParam(name: 'uploadPackage', defaultValue: false, description: 'Tick this box to upload the deb/rpm files exlude src.rpm to Artifactory')
+        booleanParam(name: 'uploadSRCRPM', defaultValue: false, description: 'Tick this box to upload the src.rpm files to Artifactory')
+        string(name: 'gitrepo', defaultValue: 'https://github.com/adoptium/installer', description: 'Do not modify unless want to build from a specific gitrepo')
+        string(name: 'gitbranch', defaultValue: 'master', description: 'Do not change unless want to build on a specific branch from above gitrepo')
+        booleanParam(name: 'enableDebug', defaultValue: false, description: 'Tick to enable --stacktrace for gradle build')
+    }
+    tools {
+        // this is set in the jenkins global tool
+        jdk "jdk-11.0.13+8"
     }
     stages {
-        stage('Linux-x64 Installers') {
-            matrix {
-                agent {
-                    label NODE_LABEL
+        stage ("Prepare Build"){
+            steps{
+                script{
+                    currentBuild.displayName =  "jdk${VERSION} - ${ARCH} - ${DISTRO}"
+                    currentBuild.description = env.BUILD_USER_ID
+                    sh("docker --version")
                 }
-                tools {
-                    jdk "jdk-11.0.13+8"
-                }
-                when {
-                    allOf {
-                        anyOf {
-                            expression { params.VERSION_FILTER == 'all' }
-                            expression { params.VERSION_FILTER == env.VERSION }
-                        }
-                        anyOf {
-                            expression { params.DISTRO_FILTER == 'all' }
-                            expression { params.DISTRO_FILTER == env.DISTRO }
-                        }
-                    }
-                }
-                axes {
-                    axis {
-                        name 'DISTRO'
-                        values 'Debian', 'RedHat', 'Suse'
-                    }
-                    axis {
-                        name 'VERSION'
-                        values '8', '11', '17', '18'
-                    }
-                }
-                stages {
-                    stage('Build Installer') {
-                        steps {
-                                echo "Installer Job for Temurin ${VERSION} - ${DISTRO}"
-                                setup()
-                                dir('linux') {
-                                    buildAndTest(DISTRO, 'x64')
-                                }
-                                script {
-                                    if (UPLOAD.toBoolean()) {
-                                        uploadArtifacts(DISTRO, 'amd64')
-                                    }
-                                }
-                            }
-                    }
+                checkout(
+                    [
+                        $class: 'GitSCM',
+                        branches: [[name: "$gitbranch"]],
+                        extensions: [[$class: 'CleanBeforeCheckout', deleteUntrackedNestedRepositories: true]],
+                        userRemoteConfigs: [[url: "$gitrepo"]]
+                    ]
+                )
+                dir("linux") {
+                    stash name: "installercode", includes: "**"
                 }
             }
         }
-        stage('Run non-x64 Deb packaging in parallel') {
-            when {
-                anyOf {
-                    expression { params.DISTRO_FILTER == 'all' }
-                    expression { params.DISTRO_FILTER == 'Debian' }
-                }
-            }
-            matrix {
-                tools {
-                    jdk "jdk-11.0.13+8"
-                }
-                when {
-                    anyOf {
-                        expression { params.VERSION_FILTER == 'all' }
-                        expression { params.VERSION_FILTER == env.VERSION }
-                    }
-                }
-                axes {
-                    axis {
-                        name 'ARCH'
-                        values 'aarch64', 'ppc64le', 's390x', 'armv7l'
-                    }
-                    axis {
-                        name 'VERSION'
-                        values '8', '11', '17', '18'
-                    }
-                }
-                excludes {
-                    exclude {
-                        axis {
-                            name 'ARCH'
-                            values 's390x'
+        stage ("BUILD")
+        {
+            parallel{
+                stage('Build Installer for Debian') {
+                    when  {
+                        beforeAgent true
+                        anyOf {
+                            expression{ params.DISTRO == 'all' }  // trigger debian build
+                            expression{ params.DISTRO == 'Debian' }
                         }
-                        axis {
-                            name 'VERSION'
-                            values '8'
+                    }
+                    agent {
+                        label "${ARCH}" == "x86_64" ? "${env.NODE_LABEL_RPM}": "${env.NODE_LABEL_DEB}"
+                    }
+                    steps{
+                        dir('linuxDebian') {
+                            script {
+                                DISTRO = "Debian"
+                                echo "Installer Job for Temurin jdk ${VERSION} - ${ARCH} - ${DISTRO}"
+                                setup("${DISTRO}", "${ARCH}")
+                                unstash 'installercode'
+                                buildAndTest("${DISTRO}", "${ARCH}")
+                                if (params.uploadPackage.toBoolean()) {
+                                    echo "Upload artifacts for ${VERSION} - ${ARCH} - ${DISTRO}"
+                                    uploadArtifacts("${DISTRO}", "${ARCH}")
+                                }
+                            }
                         }
                     }
                 }
-                agent {
-                    label "docker&&linux&&${ARCH}"
+                stage('Build Installer for Redhat') {
+                    when  {
+                        beforeAgent true
+                        anyOf {
+                            expression{ params.DISTRO == 'all' }  // trigger debian build
+                            expression{ params.DISTRO == 'RedHat' }
+                        }
+                    }
+                    steps{
+                        dir('linuxRedHat') {
+                            script {
+                                DISTRO = "RedHat"
+                                echo "Installer Job for Temurin jdk ${VERSION} - ${ARCH} - ${DISTRO}"
+                                setup("${DISTRO}", "${ARCH}")
+                                unstash 'installercode'
+                                buildAndTest("${DISTRO}", "${ARCH}")
+                                if (params.uploadPackage.toBoolean()) {
+                                    echo "Upload artifacts for ${VERSION} - ${ARCH} - ${DISTRO}"
+                                    uploadArtifacts("${DISTRO}", "${ARCH}")
+                                }
+                            }
+                        }
+                    }
                 }
-                stages {
-                    stage('Build Installer') {
-                        steps {
-                            echo "Debian Installer Job for Temurin ${VERSION} - on ${ARCH}"
-                            setup()
+                stage('Build Installer for Suse') {
+                    when  {
+                        beforeAgent true // do condition when before allocate to agent
+                        anyOf {
+                            expression{ params.DISTRO == 'all' }  // trigger debian build
+                            expression{ params.DISTRO == 'Suse' }
+                        }
+                    }
+                    steps{
+                        dir('linuxSuse') {
                             script {
-                                if (ARCH == "s390x") {
-                                    env.DOCKER_BUILDKIT=0
-                                }
-                                if (ARCH == "armv7l") {
-                                    env._JAVA_OPTIONS=""
-                                }
-                            }
-                            dir('linux') {
-                                buildAndTest('Debian', ARCH)
-                            }
-                            script {
-                                if (UPLOAD.toBoolean()) {
-                                    uploadArtifacts('Debian', ARCH)
+                                DISTRO = "Suse"
+                                echo "Installer Job for Temurin jdk ${VERSION} - ${ARCH} - ${DISTRO}"
+                                setup("${DISTRO}", "${ARCH}")
+                                unstash 'installercode'
+                                buildAndTest("${DISTRO}", "${ARCH}")
+                                if (params.uploadPackage.toBoolean()) {
+                                    echo "Upload artifacts for ${VERSION} - ${ARCH} - ${DISTRO}"
+                                    uploadArtifacts("${DISTRO}", "${ARCH}")
                                 }
                             }
                         }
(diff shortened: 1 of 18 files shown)


karianna (approved):
LGTM

linuxNew/Jenkinsfile
```diff
@@ -172,13 +172,17 @@ def uploadDebArtifacts(buildArch) {
         "s390x" : "s390x"
     ]
     def distro_list = ""
+    /*
+        debuab/Ubuntu      9.0       10.0       11.0        16.04     18.04    20.04     22.04      22.04
```
gdams, line 176:
```suggestion
        Debian/Ubuntu      9.0       10.0       11.0        16.04     18.04    20.04     22.04      22.04
```

sophia-guo (changes requested):
I'm confused. Current jenkins does support per jdk version or distro. What's the goal of this PR?

#459 requests for supporting per platform.

zdtsw:
if we are ready to run check task on non-X86 for Debian, we can take away the check on 
`def gBuildTask = (buildArch == 'x86_64') ? "packageJdk${DISTRO} checkJdk${DISTRO}" : "packageJdk${DISTRO}"`

zdtsw:
it would look something like
before:
Screenshot from 2022-06-10 16-26-30

Screenshot from 2022-06-10 16-23-52
after 
Screenshot from 2022-06-10 16-25-58

Screenshot from 2022-06-10 16-25-15

zdtsw:
> @zdtsw What is the status of this PR? If it's not suitable to be merged in its current state can it be moved into draft?

I've requested review on the last commit, but haven't got any comments on it.

gdams (approved):

karianna:
@sophia-guo left to review

linux/jdk/redhat/src/main/packaging/build.sh
```diff
@@ -4,14 +4,18 @@ set -euxo pipefail
...
+	targets="x86_64 ppc64le aarch64 armv7hl"	
+fi
+
 for spec in "$(ls /home/builder/build/generated/packaging/*.spec)"; do
 	spectool -g -R "$spec";
 	rpmbuild --nodeps -bs "$spec";
-	if [[ "$spec" =~ "temurin-8-jdk" ]]; then
-		targets="x86_64 ppc64le aarch64 armv7hl"
-	fi
 	for target in $targets; do
```
sophia-guo, line 19: A exception for targets=source should be considered?

linux/Jenkinsfile
```diff
@@ -1,126 +1,98 @@
...
+s390x       s390x       s390x       only for jdk8+
+*/
+
+env.NODE_LABEL = "dockerBuild&&linux&&x64"
+env.PRODUCT = "temurin"
 
 pipeline {
     agent {
-        label NODE_LABEL
+        label env.NODE_LABEL
```
sophia-guo, line 18: This won't work for debian non linux-x64.

sophia-guo:
Could you @zdtsw also update the PR title to correctly describe what this PR addresses?

Since the main change is with jenkinsfile the github action workflow yml may not test it ( also workflow is only for linux-x64) I would suggest you run a jenkins job to test different platforms with different distro. Feel free to use https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/ without choose upload.

sophia-guo:
While support per platform per version could we still keep the feature of support all platforms and versions? I think it's still useful especially when something wrong we can quickly and easily regenerate and upload all installers.

linux/Jenkinsfile
```diff
@@ -1,126 +1,98 @@
...
     }
     parameters {
-        string(name: 'PRODUCT', defaultValue: 'temurin')
-        string(name: 'JDK_SOURCE', defaultValue: 'Adoptium')
-        choice(name: 'VERSION_FILTER', choices: ['8', '11', '17', '18', 'all'], description: 'Build for specific VERSION')
-        choice(name: 'DISTRO_FILTER', choices: ['all', 'Debian', 'RedHat', 'Suse'], description: 'Build for specific Distro')
+        choice(name: 'VERSION', choices: ['8', '11', '17', '18'], description: 'Build for specific JDK VERSION')
+        choice(name: 'ARCH', choices: ['x86_64', 'armv7hl', 'aarch64', 'ppc64le', 's390x', 'source'], description: "Build for specific platform\n s300x not for VERSION 8\n source not for DISTRO: Debian")
+        choice(name: 'DISTRO', choices: ['all', 'Debian', 'RPM'], description: 'Build for specific Distro\n Select RPM builds for RedHat and Suse')
```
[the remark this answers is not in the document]
zdtsw: sure! if it is better we can split these two seprately. 
I was thinking if the binary passed TCK,   it is natural to have deb, and 2 rpm:s at the same time (which is the same as choose all) .  ofc, we can have a use case to only release one/two of them.
I do not know if our Jenkins has "Multiselector Parameter" plugin installed or not. if that is done, we can  have DISTRO: `[ 'Debian', 'RPM', 'Suse']` which allows us to choose three distro (same as '`all`' option) or just one of them

linux/jdk/redhat/src/main/packaging/build.sh
zdtsw: yes, 'source' should be added

zdtsw:
> While support per platform per version could we still keep the feature of support all platforms and versions? I think it's still useful especially when something wrong we can quickly and easily regenerate and upload all installers.

I am thinking to keep both as two different jobs, it is up to the one doing release to see which fits better per each case.

zdtsw:
> Could you @zdtsw also update the PR title to correctly describe what this PR addresses?
> 
> Since the main change is with jenkinsfile the github action workflow yml may not test it ( also workflow is only for linux-x64) I would suggest you run a jenkins job to test different platforms with different distro. Feel free to use https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/ without choose upload.

thanks! will do

linux/jdk/redhat/src/main/packaging/build.sh
zdtsw: I looked up the change, in the jenkins dropdown i have this source, but the code it does not need source as input
because i does the` rpmbuild -bs `first then do the real `rpmbuild -target `
so we actually do not need an option in jenkins to choose for `source`

linux/Jenkinsfile
zdtsw: good catch! I missed it

zdtsw:
by checking up the failed run, I realized the problem was how we design these tests in the past.

for RPM package test, it assumes that should be a x86_64.rpm available among other rpms. 
when this is changed (when i was test on suse + non-x86_64) test failed because we only generate a `*.src.rpm` + `*.armv7hl.rpm`
therefore the problem was not only on Suse but both RPM jobs.

zdtsw:
another change in the junit is to only test "yum/dnf/zypper install" when ARCH is set to x86_64
for another other arch, test covers if rpm package has been successfully built.

zdtsw:
RedHat arm32:
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/125/console

RedHat x64:
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/129/console



Suse x86:
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/127/console

Suse aarch64
https://ci.adoptopenjdk.net/view/work-in-progress/job/Sophia-adoptium-packages-linux-pipeline/128/console

zdtsw:
need another round of review on this PR

most of the big changes have been caught  in the PR description. 

and test result is updated with links too.
GH action covers the ARCH=="all" scenario

zdtsw:
@steelhead31 maybe you can do a review on the` dpkg-buildpackage` part ?  with all these flags etc.

steelhead31 (approved):
The dpkg changes look good to me.

sxa (commented):
I'm going to let others review this properly as I'm not familiar enough with the code to do a suitable job on it given the extent of the changes. We already have the ability to run per version I believe, so for me the important bit about this PR is the ability to run per-architecture as per https://github.com/adoptium/installer/issues/459

It looks like this now needs a rebase, and I'd suggest running a few tests through jenkins on the latest code to verify that it is able to support that use case properly on each platform since the links in the description are now obsolete.

linux/Jenkinsfile
```diff
@@ -1,126 +1,133 @@
...
+        timeout(time: 2, unit: 'HOURS') 
     }
     parameters {
-        string(name: 'PRODUCT', defaultValue: 'temurin')
-        string(name: 'JDK_SOURCE', defaultValue: 'Adoptium')
-        choice(name: 'VERSION_FILTER', choices: ['8', '11', '17', '18', 'all'], description: 'Build for specific VERSION')
-        choice(name: 'DISTRO_FILTER', choices: ['all', 'Debian', 'RedHat', 'Suse'], description: 'Build for specific Distro')
-        booleanParam(name: 'UPLOAD', defaultValue: false, description: 'Tick this box to upload the deb/rpm files to Artifactory')
+        choice(name: 'VERSION', choices: ['8', '11', '17', '18'], description: 'Build for specific JDK VERSION')
+        choice(name: 'ARCH', choices: ['x86_64', 'armv7hl', 'aarch64', 'ppc64le', 's390x', 'all'], description: "Build for specific platform\n s300x not for VERSION 8\n")
```
sophia-guo, line 26: arm should be `armv7l`

linux/Jenkinsfile
```diff
@@ -131,82 +138,106 @@ pipeline {
...
+*/
+def setup(DISTRO, buildArch) {
     cleanWs()
     // Docker --mount option requires BuildKit
-    env.DOCKER_BUILDKIT=1
+    env.DOCKER_BUILDKIT = buildArch == "s390x" ? '0' : '1'
     env.COMPOSE_DOCKER_CLI_BUILD=1
-    env._JAVA_OPTIONS="-Xmx4g"
-    checkout scm
+    env._JAVA_OPTIONS =  (buildArch == "armv7hl" && DISTRO == "Debian")? "" : "-Xmx4g"
```
sophia-guo, line 149: buildArch=`armv7l`

linux/Jenkinsfile
```diff
@@ -131,82 +138,106 @@ pipeline {
...
 }
 
 def uploadDebArtifacts(buildArch) {
-    def archs = [
-        "amd64" : "amd64",
-        "ppc64le": "ppc64el",
+    // full list of all platforms, up to user to opt out s390x+jdk8
+    def debArchList = [
+        "x86_64" : "amd64",
+        "armv7hl": "armhf",
```
sophia-guo, line 194: `armv7l`

linux/Jenkinsfile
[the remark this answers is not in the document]
zdtsw: I think with the final commit, this split has been done

linux/Jenkinsfile
```diff
@@ -1,126 +1,98 @@
...
-                        anyOf {
-                            expression { params.VERSION_FILTER == 'all' }
-                            expression { params.VERSION_FILTER == env.VERSION }
-                        }
+            }
+        }
+        stage ("BUILD")
+        {
+            parallel{
+                stage('Build Installer for Debian') {
```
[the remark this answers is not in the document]
zdtsw: this should also be resolve when apply `NODE_LABEL_DEB` with `${ARCH}` as part of label for deb job.

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
+        choice(name: 'DISTRO', choices: ['all', 'Debian', 'RedHat', 'Suse'], description: 'Build for specific Distro\n Select RPM builds for RedHat and Suse')
+        booleanParam(name: 'uploadPackage', defaultValue: false, description: 'Tick this box to upload the deb/rpm files exlude src.rpm to Artifactory')
+        booleanParam(name: 'uploadSRCRPM', defaultValue: false, description: 'Tick this box to upload the src.rpm files to Artifactory')
+        string(name: 'gitrepo', defaultValue: 'https://github.com/adoptium/installer', description: 'Do not modify unless want to build from a specific gitrepo')
+        string(name: 'gitbranch', defaultValue: 'master', description: 'Do not change unless want to build on a specific branch from above gitrepo')
+        booleanParam(name: 'enableDebug', defaultValue: false, description: 'Tick to enable --stacktrace for gradle build')
+    }
+    tools {
+        // this is set in the jenkins global tool
+        jdk "jdk-11.0.13+8"
```
karianna, line 36: Should try to get this to 11.0.16.1

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
-        stage('Linux-x64 Installers') {
-            matrix {
-                agent {
-                    label NODE_LABEL
+        stage ("Prepare Build"){
+            steps{
+                script{
+                    currentBuild.displayName =  "jdk${VERSION} - ${ARCH} - ${DISTRO}"
+                    currentBuild.description = env.BUILD_USER_ID
+                    sh("docker --version")
```
karianna, line 44: Is that just for debug?

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
+        stage ("BUILD")
+        {
+            parallel{
+                stage('Build Installer for Debian') {
+                    when  {
+                        beforeAgent true
                         anyOf {
-                            expression { params.DISTRO_FILTER == 'all' }
-                            expression { params.DISTRO_FILTER == env.DISTRO }
+                            expression{ params.DISTRO == 'all' }  // trigger debian build
```
karianna, line 66: comment on the wrong line?

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
-                            values 's390x'
-                        }
-                        axis {
-                            name 'VERSION'
-                            values '8'
+                stage('Build Installer for Redhat') {
+                    when  {
+                        beforeAgent true
+                        anyOf {
+                            expression{ params.DISTRO == 'all' }  // trigger debian build
```
karianna, line 93: wring comment on wrong line :-)

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
-                        steps {
-                            echo "Debian Installer Job for Temurin ${VERSION} - on ${ARCH}"
-                            setup()
+                    steps{
+                        dir('linuxRedHat') {
                             script {
-                                if (ARCH == "s390x") {
-                                    env.DOCKER_BUILDKIT=0
+                                DISTRO = "RedHat"
+                                echo "Installer Job for Temurin jdk ${VERSION} - ${ARCH} - ${DISTRO}"
```
karianna, line 101: This seems to be repeating similar code to above, can it be extracted into a function?

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
-                                buildAndTest('Debian', ARCH)
                             }
+                        }
+                    }
+                }
+                stage('Build Installer for Suse') {
+                    when  {
+                        beforeAgent true // do condition when before allocate to agent
+                        anyOf {
+                            expression{ params.DISTRO == 'all' }  // trigger debian build
```
karianna, line 117: wrong comment on wrong line :-)

linux/Jenkinsfile
```diff
@@ -1,126 +1,134 @@
...
-                            }
-                            dir('linux') {
-                                buildAndTest('Debian', ARCH)
                             }
+                        }
+                    }
+                }
+                stage('Build Installer for Suse') {
+                    when  {
+                        beforeAgent true // do condition when before allocate to agent
```
karianna, line 115:
```suggestion
                        beforeAgent true // execute the following block before we allocate the agent
```

linux/Jenkinsfile
```diff
@@ -131,69 +139,92 @@ pipeline {
...
 }
 
-def buildAndTest(DISTRO, arch) {
-    if (DISTRO != "Debian") {
-        // Install Adoptium GPG key for RPM signing
-        withCredentials([file(credentialsId: 'adoptium-artifactory-gpg-key', variable: 'GPG_KEY')]) {
-            sh('./gradlew packageJdk${DISTRO} checkJdk${DISTRO} --parallel -PPRODUCT=${PRODUCT} -PPRODUCT_VERSION=${VERSION} -PGPG_KEY=${GPG_KEY}')
+def buildAndTest(DISTRO, buildArch) {
+    try {
+        if (DISTRO != "Debian") { // for RPM based: RedHat / Suse
```
karianna, line 155: This might be brittle later, better to explicity say if Suse or if REdHat

linux/Jenkinsfile
```diff
@@ -131,69 +139,92 @@ pipeline {
...
 
-def uploadArtifacts(distro, arch) {
-    if (distro == "Debian") {
-        uploadDebArtifacts(arch)
-    } else {
-        uploadRPMArtifacts(distro)
+def uploadArtifacts(DISTRO, buildArch) {
+    if (DISTRO == "Debian") {
+        uploadDebArtifacts(buildArch)
+    } else {  // DISTRO == RPM(RedHat||Suse)
```
karianna, line 186: could be slightly brittle, maybe have an else if for this and then and else at the end which errors if distro is not valid

linux/Jenkinsfile
```diff
@@ -131,69 +139,92 @@ pipeline {
...
+        uploadRPMArtifacts(DISTRO)
     }
 }
 
 def uploadDebArtifacts(buildArch) {
-    def archs = [
-        "amd64" : "amd64",
-        "ppc64le": "ppc64el",
-        "aarch64": "arm64",
+    // full list of all platforms, up to user to opt out s390x+jdk8
```
karianna, line 192: comment doesn't match the code?

linux/README.md
```diff
@@ -160,9 +160,9 @@ rpmbuild --define "_sourcedir $(pwd)" --define "_specdir $(pwd)" \
 ## Supported packages
 
 ### DEB
-Supported JDK version 8,11,17,18 
+Supported JDK version 8,11,17,18
 
-Supported platform amd64, arm64, armhf, ppc64le, s390x (s390x is only available for jdk11+)  
+Supported platform amd64, arm64, armhf, ppc64le, s390x (s390x is only available for jdk8+)
```
karianna, line 165: not sure that comment is correct, should this be `>8`?

linux/README.md
```diff
@@ -178,7 +178,7 @@ Supported platform amd64, arm64, armhf, ppc64le, s390x (s390x is only available
 ### RPM (RedHat and Suse)
 Supported JDK version 8,11,17,18
 
-Supported platform x86_64, aarch64, armv7hl, ppc64le, s390x (s390x is only available for jdk11+)
+Supported platform x86_64, aarch64, armv7hl, ppc64le, s390x (s390x is only available for jdk8+)
```
karianna, line 181: not sure that comment is correct, should this be `>8`?

linux/jdk/debian/src/main/packaging/build.sh
```diff
@@ -6,11 +6,22 @@ set -euxo pipefail
...
-# Build package and set distributions it supports.
-# $VERSIONS are container env. variables
+
+# $ and $ARCH are env variables passing in from "docker run"
+debVersionList="buster bullseye bionic focal jammy"
+dpkgExtraARG="-us -uc" # ignore building with a gpg key
+
+echo "DEBUG: building Debian arch ${buildArch}"
+if [[ "${buildArch}" == "all" ]]; then
+	dpkgExtraARG="${dpkgExtraARG} -b" # equal to --build=any,all|--build=binary
```
karianna, line 16:
```suggestion
    dpkgExtraARG="${dpkgExtraARG} -b" # equal to --build=any,all|--build=binary
```

linux/jdk/redhat/src/packageTest/java/packaging/DnfOperationsTest.java
```diff
@@ -53,27 +53,30 @@ void packageSuccessfullyInstalled(String distribution, String codename) throws E
...
-					.contains("Packager    : Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>");
-			} else {
-				assertThat(result.getStdout())
-					.contains("Name        : " + System.getenv("PACKAGE"))
-					.contains("Group       : java")
-					.contains("License     : GPLv2 with exceptions")
-					.contains("Signature   : (none)")
-					.contains("Relocations : /usr/lib/jvm/" + System.getenv("PACKAGE"))
-					.contains("Packager    : Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>");
+			// below part: only test x86_64 rpm package in docker container
```
karianna, line 56: comment doesn't match the code, wouldn't all architectures include aarch64 etc?

linux/jdk/redhat/src/packageTest/java/packaging/RpmFiles.java
```diff
@@ -34,7 +35,14 @@ private RpmFiles() {
 	}
 
 	static Path hostRpmPath() {
-		return findBuildOutputOnHost(System.getenv("PACKAGE") + "*.x86_64.rpm");
+		// convert filter when build with ARCH == all
+		Map<String, String> env  = System.getenv(); 
+        for (String envName : env.keySet()) {
```
karianna, line 40: whitespace seems incorrect (needs indenting)

linux/jdk/suse/build.gradle
```diff
@@ -3,9 +3,9 @@ plugins {
 }
```
karianna, line 3: seems to be some common functionality in these scripts, can we extract?

karianna (changes requested):

linux/Jenkinsfile
zdtsw: if 11.0.16.1 is available on ci.adoptopenjdk, i can uplift the version

zdtsw: Screenshot from 2022-09-08 11-51-21
only see 11.0.13+8

linux/Jenkinsfile
zdtsw: the testcontiners use the docker installed on the agent, to print docker version so we know which we use in the test stage.

linux/Jenkinsfile
zdtsw: seems got copied up on all cases...

linux/README.md
zdtsw: is 8+ == >8 :D 
i think, jdk11+ means, starting from 12 and so on.
at least we do have s390x for jdk11

linux/jdk/redhat/src/packageTest/java/packaging/DnfOperationsTest.java
zdtsw: so, when explicit set `ARCH` to `x86_64` or all or do not set ARCH at all , this does tests on these asserts
at the moment (before this PR) we only have "all" so it does build x86_64+ s390+ppc64le etcetc, but the test only done for the x86_64 one, the rest are not tested , which is the same as `System.getenv("testArch") == "all"`

linux/jdk/suse/build.gradle
zdtsw: it is quite interesting. 
most of the config in gradle already defined in the root project (linux/build.gradle), but then in different distro they are defined again. guess it is for overwritten purpose when it was tested. 
these lines of code should be able to remove

linux/Jenkinsfile
zdtsw: i must correct myself. apart from testcontainers, we also run "docker build" "docker run" from gradle, thats another reason want to have docker version print out

linux/README.md
karianna, line 181:
```suggestion
Supported platform x86_64, aarch64, armv7hl, ppc64le, s390x (s390x is only available for jdk > 8)
```

linux/Jenkinsfile
zdtsw: move into function` jenkinsStep()`

linux/jdk/suse/build.gradle
zdtsw: i am thinking to leave the "code improvement" into another issue, so we can get review done for the current change first

linux/Jenkinsfile
zdtsw: did test we need 11.0.16.1 installed as global "tools" before we can set to use it
```
WorkflowScript: 14: Tool type "jdk" does not have an install of "11.0.16.1" configured - did you mean "jdk-11.0.13+8"? @ line 14, column 13.
           jdk "11.0.16.1"
```

linux/Jenkinsfile
```diff
@@ -214,22 +235,22 @@ def uploadRPMArtifacts(distro) {
...
+        "source" : "src",  // here we need src in the list to be able to upload
+        "s390x"  : "s390x"
     ]
-    def packageDirs = distro_Package.get(tempDistro)
-    
-    if ( !VERSION.equals("8") ) {
-        archs.put("s390x", "s390x")
-    }
+    // Enable upload src.rpm
+    if ( params.uploadSRCRPM.toBoolean() || params.DISTRO == 'all'){   rpmArchList.put("Source","src") }
```
sophia-guo, line 249: same as other item prefer "Source" => "source"

linux/jdk/redhat/src/packageTest/java/packaging/RedHatFlavoursWithYum.java
```diff
@@ -46,7 +46,9 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
 			Arguments.of("amazonlinux", "2"),
 			Arguments.of("centos", "7"),
 			Arguments.of("registry.access.redhat.com/ubi7/ubi", "latest"),
-			Arguments.of("oraclelinux", "7")
+			Arguments.of("redhat/ubi8", "latest"),
+			Arguments.of("oraclelinux", "7"),
+			Arguments.of("oraclelinux", "8")
```
sophia-guo, line 51: Can also add 'redhat/ubi9' for  'rpm/rhel/9'?

zdtsw: in theory, we do not need to test the same OS for both yum and dnf. some old ones, e.g ubi7 does not support dnf so it has to be in  yum test

I intended to remove oraclelinux 8 and ubi8 from Yum as well (already covered in dnf)

sophia-guo (approved):

linux/Jenkinsfile
karianna: We can make this change later

karianna (approved):

PR #474, opened 2022-05-24
refact: move build.sh as file not generate everytime in docker build

zdtsw:
- for better practice and to reduce maintenance work: create build.sh as static file not generate as step in docker build 
	- install tini from dnf not download from github
	- remove install "sudo" in dockerfile

### DIFF
--- a/linux/ca-certificates/debian/build.gradle
+++ b/linux/ca-certificates/debian/build.gradle
@@ -61,7 +61,7 @@ task packageDebian {
 	dependsOn "assemble"
 
 	group = "packaging"
-	description = "Creates deb package for Debian flavours."
+	description = "Creates ca-certificate deb package for Debian flavours."
 
 	def outputDir = new File(project.buildDir.absolutePath, "ospackage")
 	outputs.dir(outputDir)
@@ -93,7 +93,7 @@ task packageDebian {
 task checkDebianPackage(type: Test) {
 	dependsOn packageDebian
 
-	description = 'Tests the generated Debian packages.'
+	description = 'Tests the generated ca-certificate Debian packages.'
 	group = 'verification'
 
 	testClassesDirs = sourceSets.packageTest.output.classesDirs
--- a/linux/ca-certificates/debian/src/main/packaging/Dockerfile
+++ b/linux/ca-certificates/debian/src/main/packaging/Dockerfile
@@ -2,32 +2,16 @@ FROM debian:buster
 
 # Combine apt-get update with apt-get install to prevent stale package indexes.
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential \
- 		debhelper sudo lsb-release reprepro gosu tini
+ 		debhelper lsb-release reprepro gosu tini
 
 # Create unprivileged user for building, see
 # https://github.com/hexops/dockerfile#use-a-static-uid-and-gid
-RUN groupadd -g 10001 builder
-RUN useradd -m -d /home/builder -u 10000 -g 10001 builder
+RUN groupadd -g 10001 builder \
+	&& useradd -m -d /home/builder -u 10000 -g 10001 builder
 
 # Prepare entrypoint and build scripts
 ADD entrypoint.sh /entrypoint.sh
-RUN printf '#!/usr/bin/env bash\n\
-set -euxo pipefail\n\
-\n\
-# Copy build scripts into a directory within the container. Avoids polluting the mounted\n\
-# directory and permission errors.\n\
-mkdir /home/builder/workspace\n\
-cp -R /home/builder/build/generated/packaging /home/builder/workspace\n\
-\n\
-# Build package and set distributions it supports.\n\
-cd /home/builder/workspace/packaging\n\
-dpkg-buildpackage -us -uc -b\n\
-changestool /home/builder/workspace/*.changes setdistribution $VERSIONS\n\
-\n\
-# Copy resulting files into mounted directory where artifacts should be placed.\n\
-mv /home/builder/workspace/*.{deb,changes,buildinfo} /home/builder/out\n\
-'\
->> /home/builder/build.sh
+ADD build.sh /home/builder/build.sh
 RUN chmod +x /home/builder/build.sh
 
 ENTRYPOINT ["/usr/bin/tini", "--", "/bin/bash", "/entrypoint.sh" ]
--- a/linux/ca-certificates/debian/src/main/packaging/build.sh
+++ b/linux/ca-certificates/debian/src/main/packaging/build.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+# Copy build scripts into a directory within the container. Avoids polluting the mounted
+# directory and permission errors
+mkdir /home/builder/workspace
+cp -R /home/builder/build/generated/packaging /home/builder/workspace
+
+# Build package and set distributions it supports
+cd /home/builder/workspace/packaging
+dpkg-buildpackage -us -uc -b
+changestool /home/builder/workspace/*.changes setdistribution $VERSIONS
+
+# Copy resulting files into mounted directory where artifacts should be placed.
+mv /home/builder/workspace/*.{deb,changes,buildinfo} /home/builder/out
--- a/linux/gradlew
+++ b/linux/gradlew
@@ -126,7 +126,7 @@ if $darwin; then
     GRADLE_OPTS="$GRADLE_OPTS \"-Xdock:name=$APP_NAME\" \"-Xdock:icon=$APP_HOME/media/gradle.icns\""
 fi
 
-# For Cygwin or MSYS, switch paths to Windows format before running java
+# For Cygwin or MSYS/mingw, switch paths to Windows format before running java
 if [ "$cygwin" = "true" -o "$msys" = "true" ] ; then
     APP_HOME=`cygpath --path --mixed "$APP_HOME"`
     CLASSPATH=`cygpath --path --mixed "$CLASSPATH"`
--- a/linux/jdk/debian/src/main/packaging/Dockerfile
+++ b/linux/jdk/debian/src/main/packaging/Dockerfile
@@ -2,33 +2,32 @@ FROM debian:buster
 
 # Combine apt-get update with apt-get install to prevent stale package indexes.
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential \
- 		debhelper sudo lsb-release reprepro gosu java-common libasound2 libc6 libx11-6 \
- 		libfontconfig1 libfreetype6 libxext6 libxi6 libxrender1 libxtst6 zlib1g tini wget
+		debhelper \
+		lsb-release \
+		reprepro \
+		gosu \
+		java-common \
+		libasound2 \
+		libc6 \
+		libx11-6 \
+		libfontconfig1 \
+		libfreetype6 \
+		libxext6 \
+		libxi6 \
+		libxrender1 \
+		libxtst6 \
+		zlib1g \
+		tini \
+		wget
 
 # Create unprivileged user for building, see
 # https://github.com/hexops/dockerfile#use-a-static-uid-and-gid
-RUN groupadd -g 10001 builder
-RUN useradd -m -d /home/builder -u 10000 -g 10001 builder
+RUN groupadd -g 10001 builder \
+	&& useradd -m -d /home/builder -u 10000 -g 10001 builder
 
 # Prepare entrypoint and build scripts
 ADD entrypoint.sh /entrypoint.sh
-RUN printf '#!/usr/bin/env bash\n\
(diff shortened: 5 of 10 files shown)


linuxNew/README.md
```diff
@@ -114,11 +123,13 @@ rpm-build-4.16.1.3-1.fc33.x86_64
 
 Consider this RPM build where x86_64 is the build hosts' architecture.
 Download the release blobs and associated sources.
+Suppose build rpm for jdk11 for target architecture `x86_64`
 
 ```shell
-spec=$(pwd)/temurin-11-jdk.spec
+cd linuxNew/jdk/redhat/src/main/packaging/temurin/11
```
karianna, line 129: Won't be linuxNew much longer I suppose

README.md
```diff
@@ -2,70 +2,60 @@
 Repository for creating installable packages for AdoptOpenJDK releases.
 
 The packages are created using:
-1. The Wix Toolset http://wixtoolset.org (Windows only)
-2. [Packages](http://s.sudre.free.fr/Software/Packages/about.html) (Mac OS)
-3. For putting together `.deb` and `rpms` head to this link: [linuxNew subdir readme](https://github.com/adoptium/installer/tree/master/linuxNew#readme)
+1. [Wix Toolset](http://wixtoolset.org) (Windows), detail see [implementation](./wix)
+2. [Packages](http://s.sudre.free.fr/Software/Packages/about.html) (MacOS), detail see (implementation](./pkgbuild)
+3.  Linux installer include deb and rpm based package, detail see [implementation](./linuxNew)
```
karianna, line 7: change to linux

linuxNew/README.md
```diff
@@ -24,14 +24,14 @@ Builds take at least ~5-15 minutes to complete on a modern machine.  Please ensu
 
 You'll want to make sure you've set the exact versions of the binaries you want package in the:
 
-* **Debian Based** - _jdk\debian\src\main\packaging\\&lt;vendor&gt;\\&lt;version&gt;\\debian\\rules_ files.
-* **Red Hat Based** - _jdk\redhat\src\main\packaging\\&lt;vendor&gt;\\&lt;version&gt;\\&lt;vendor&gt;-&lt;version&gt;-jdk.spec_ files.
-* **SUSE Based** - _jdk\suse\src\main\packaging\\&lt;vendor&gt;\\&lt;version&gt;\\&lt;vendor&gt;-&lt;version&gt;-jdk.spec_ files.
+*  **Debian Based** - _jdk/debian/src/main/packaging/\<vendor>\/\<version>\/debian/rules_ files.
```
karianna, line 27: Extra whitespace not needed?

README.md
```diff
@@ -2,70 +2,60 @@
 Repository for creating installable packages for AdoptOpenJDK releases.
```
karianna, line 2: AdoptOpenJDK --> Eclipse Adoptium based

karianna (changes requested):

linuxNew/README.md
zdtsw: yeah, PR of renaming "linuxNew" to "linux" need rebase this PR first and updates these references.

README.md
zdtsw: this is a hyperlink, if we keep folder "linuxNew" for now and update it to (./linux) it actually creates a broken link. 
better to keep it for now, and update it when we finalize  the renaming PR

karianna (changes requested):
LGTM but can you restore the azure devops file - we (MSFT) intend to revive that work

karianna (approved):

README.md
```diff
@@ -1,71 +1,61 @@
...
-      - TAG: jdk-16+36
-      - SUB_TAG: 16_36
-3. After each Jenkins job verify the success of the job, and also verify the artifacts have been uploaded to Artifactory by checking
-this location:
-  - https://adoptopenjdk.jfrog.io/ui/repos/tree/General/deb%2Fpool%2Fmain%2Fa
-    - Verify the artifacts exist under the relevant sub-folder for the job run, eg.adoptopenjdk-11-hotspot, for jdk-11 hotspot JDK, and
-adoptopenjdk-11-hotspot-jre for jdk-11 hotspot JRE
-
+## Releasing Mac and Windows Installer packages
+This is done from GitHub Action
```
sxa, line 16: This is incorrect. They are created during the build pipelines at https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/create_installer_mac/ and https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/create_installer_windows/

README.md
```diff
@@ -1,71 +1,61 @@
...
+## Releasing Linux Installer packages
+During a Release the Linux installers (deb,rpm) are not created as part of the build job, but are instead created manually after the production binaries have been published to https://github.com/adoptium/temurinXX-binaries/releases.
+The following documentation describes how to create and publish these Linux installers to [Artifactory](https://adoptium.jfrog.io/ui/repos/tree/General)
+
+1. Check the given jdk version binaries have been published to GitHub, "latest" should be for Hotspot:
+  - jdk8 : https://github.com/adoptium/temurin8-binaries/releases/latest
+  - jdk11 : https://github.com/adoptium/temurin11-binaries/releases/latest
+  - jdk17 : https://github.com/adoptium/temurin17-binaries/releases/latest
+  - jdk18 : https://github.com/adoptium/temurin18-binaries/releases/latest
+2. For each jdk version and JVM variant, run the following [Jenkins job](https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/standalone_create_installer_linux) (Restricted auth permission) to create and publish Linux installers to Artifactory:
```
sxa, line 27: Most of this information is incorrect - that job is no longer used since we moved to Adoptium, and the correct details will be in the documentation in the linuxNew directory.

linuxNew/jdk/redhat/src/main/packaging/Dockerfile
```diff
@@ -1,25 +1,24 @@
 FROM fedora:33
 
-RUN dnf update -y && dnf install -y rpmdevtools rpm-sign rpmlint gnupg2 wget tar dpkg findutils
+RUN dnf update -y && dnf install -y rpmdevtools rpm-sign rpmlint \
```
sxa, line 3: Any particular reason why the first 3 are on one line then each of the others are on their own? Seems inconsistent.

linuxNew/jdk/redhat/src/main/packaging/Dockerfile
```diff
@@ -32,37 +31,11 @@ RUN --mount=type=secret,id=gpg,gid=10001,uid=10000,dst=/tmp/private.gpg \
...
-# Copy generated SRPMS, RPMs to destination folder\n\
-find /home/builder/rpmbuild/SRPMS /home/builder/rpmbuild/RPMS -type f -name "*.rpm" -print0 | xargs -0 -I {} cp {} /home/builder/out\n\
-# Sign generated RPMs with rpmsign.\n\
-if grep -q %%_gpg_name /home/builder/.rpmmacros; then \n\
-	rpmsign --addsign /home/builder/out/*.rpm \n\
-fi;\n\
-'\
->> /home/builder/build.sh
+ADD build.sh /home/builder/build.sh
+USER root
```
sxa, line 38: Why has this been added explicitly?

linuxNew/jdk/suse/src/main/packaging/Dockerfile
```diff
@@ -25,38 +31,11 @@ RUN --mount=type=secret,id=gpg,gid=10001,uid=10000,dst=/tmp/private.gpg \
...
-# Copy generated RPMs to destination folder\n\
-find /home/builder/rpmbuild/SRPMS /home/builder/rpmbuild/RPMS -type f -name "*.rpm" -print0 | xargs -0 -I {} cp {} /home/builder/out\n\
-# Sign generated RPMs with rpmsign.\n\
-if grep -q %%_gpg_name /home/builder/.rpmmacros; then \n\
-	rpmsign --addsign /home/builder/out/*.rpm \n\
-fi;\n\
-'\
->> /home/builder/build.sh
+ADD build.sh /home/builder/build.sh
+USER root
```
sxa, line 38: Same comment as to why this has been added

linuxNew/jdk/redhat/src/main/packaging/Dockerfile
zdtsw: do not want to keep the line too long.
can make each package one line instead

linuxNew/jdk/redhat/src/main/packaging/Dockerfile
zdtsw: from https://github.com/adoptium/installer/blob/master/linuxNew/jdk/redhat/src/main/packaging/Dockerfile#L35
we use "root" to perform entrypoint.
so this is not added in the change.

and  this change, in order to run "chmod +x" for build.sh
otherwise, permission denied. 
see log https://github.com/adoptium/installer/runs/6656938909?check_suite_focus=true

` 
RUN chmod +x /home/builder/build.sh:  
executor failed running [/bin/sh -c chmod +x /home/builder/build.sh]: exit code: 1`

linuxNew/jdk/suse/src/main/packaging/Dockerfile
zdtsw: same as above, not added by this PR

README.md
zdtsw: Good to know! 
I still have no permission to access "release" can be wrong some parts that invisible for me.

I agree, maybe this PR should only include the non-doc change. since some part I dont have permission, can give the wrong information +  can wait till renaming linuxNew to linux done

linuxNew/jdk/redhat/src/main/packaging/Dockerfile
sxa: Do you know how it worked before the change? The `chmod` isn't new in here.

linuxNew/jdk/suse/src/main/packaging/Dockerfile
sxa: Ah I've just seen that it's just been moved to a different place in the file. So it's now after the addition of the entrypoint and build.sh files and no longer owned by the `builder` ID

linuxNew/jdk/redhat/src/main/packaging/Dockerfile
sxa: While I'm still not sure why the `USER root` statement was moved it doesn't negatively impact the functionality so I'm ok with it now.

sophia-guo:
what's the benefit of creating build.sh as static file not generate as step in docker build? All paths and tools in generated build.sh are specific to Dockerfile image environment. IMO it doesn't make sense to add those temporary files in repo.

zdtsw:
The reason to move these into a seprate script file is for readibility. This could make the code easier to maintain for others in the future. If someday we need to add more logic into this block, it would be much easy to update one file, and test is offline instead of doing "docker build" everytime to test script.

Moreover, decouple logic from Dockerfile.  
Dockerfile is supposed to only handle build steps, not to use to generated logic.

We already ADD entrypoint.sh, by doing can keep consistent of how we handle scripts.

It is still kept  different build.sh per each "distro". In reality the one used by Red Hat and Suse can be share (but this is not in the scope of this PR) 

As for env, it does not happen in build time, only run time requires env variable (e.g VERSIONS) this is passed by as "-e" from "docker run" either from gradle build in Jenkins or GH action step.

Dockerfile has already been stored in git, it should not be a big concern to store temporary file in git.

One gain by doing this is to have a better understading to see what has been done when built the image.
Running "docker history" on final imag, 
shows step "ADD build.sh /tmp/build.sh # buildkit"
than previous "RUN /bin/sh -c printf '#!/usr/bin/env bash\n…  "

README.md
```diff
@@ -1,71 +1,61 @@
...
-adoptopenjdk-11-hotspot-jre for jdk-11 hotspot JRE
-
+## Releasing Mac and Windows Installer packages
+This is done from GitHub Action
 
+## Releasing Linux Installer packages
+During a Release the Linux installers (deb,rpm) are not created as part of the build job, but are instead created manually after the production binaries have been published to https://github.com/adoptium/temurinXX-binaries/releases.
+The following documentation describes how to create and publish these Linux installers to [Artifactory](https://adoptium.jfrog.io/ui/repos/tree/General)
+
+1. Check the given jdk version binaries have been published to GitHub, "latest" should be for Hotspot:
```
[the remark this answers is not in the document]
sxa: Resolving as this doc change has disappeared from the PR now I believe

README.md
zdtsw: once https://github.com/adoptium/installer/pull/478 get merged(today hopefully) will do a rebase, these should be updated accordingly

sxa (approved):

README.md
karianna: OK, I've landed #478 so you can rebase now.

zdtsw: thanks!

karianna:
@zdtsw Will need a rebase now

PR #475, opened 2022-05-24
refact: cleanup "linux" folder which is not in use by Adoptium

zdtsw:


karianna (approved):
LGTM

karianna:
Maybe drop a note in the installers channel in Slack and warn folks this is going away.  Then give it 24 hours or so before merging

zdtsw:
think we are ready to  have this merged? @karianna

PR #486, opened 2022-06-23
Remove alternatives during prerm phase

luozhenyu:
# Unexpected behavior
After executing `apt remove temurin-8-jdk`, the `update-alternatives --remove` will not be executed.

# Cause
1. The existence of `tool_path` will be checked before `update-alternatives --remove` executes.
2. However, the `tool_path` does not exist in the `postrm` phase. 

# Solution
1. Execute `update-alternatives --remove` in the `prerm` phase

https://github.com/adoptium/installer/blob/master/linux/jdk/debian/src/main/packaging/temurin/8/debian/postrm.in#L14-L18


luozhenyu:
@karianna This is my first commit and I signed the Eclipse CLA just now.

karianna (approved):
changes LGTM

PR #487, opened 2022-06-23
Correct the priority of temurin-17-jdk and temurin-18-jdk & Remove jaotc and rmid from jvm tools.

luozhenyu:
1. Change the priority of temurin-17-jdk to [1711](https://salsa.debian.org/openjdk-team/openjdk/-/blob/openjdk-17/debian/rules#L96) 
2. Change the priority of temurin-18-jdk to [1811](https://salsa.debian.org/openjdk-team/openjdk/-/blob/openjdk-18/debian/rules#L96)
3. Remove jaotc and rmid from jvm tools (JDK 17/18)
   1. [JEP 407: Remove RMI Activation](https://openjdk.org/jeps/407)
   2. [JEP 410: Remove the Experimental AOT and JIT Compiler](https://openjdk.org/jeps/410)

### DIFF
--- a/linux/jdk/debian/src/main/packaging/temurin/17/debian/rules
+++ b/linux/jdk/debian/src/main/packaging/temurin/17/debian/rules
@@ -2,7 +2,7 @@
 
 pkg_name = temurin-17-jdk
 priority = 1711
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM

PR #491, opened 2022-06-26
Documentation: Docker 20.10+ is required to build

luozhenyu:
# Changes
1. Docker 20.10+ is required to build.
    1. Because the `--mount` option is used without specify experimental dockerfile `#syntax` directive. https://github.com/adoptium/installer/blob/master/linux/jdk/redhat/src/main/packaging/Dockerfile#L27
    > https://docs.docker.com/engine/release-notes/#20100
    > buildkit,dockerfile: Support for RUN --mount options without needing to specify experimental dockerfile #syntax directive. moby/buildkit#1717
2. Remove `COMPOSE_DOCKER_CLI_BUILD`
    1. Because docker compose is not used.
    2. Because `COMPOSE_DOCKER_CLI_BUILD=1` is the default setting. https://docs.docker.com/compose/reference/envvars/#compose_docker_cli_build

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -73,7 +73,6 @@ jobs:
         run: |
           export _JAVA_OPTIONS="-Xmx4G"
           export DOCKER_BUILDKIT=1
-          export COMPOSE_DOCKER_CLI_BUILD=1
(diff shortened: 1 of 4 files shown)


zdtsw:
Sure I agree we should remove unnecessary flag in our code to not cause confusion `COMPOSE_DOCKER_CLI_BUILD`

As for specify mini docker version, I am not aware of anyone doing local package build. 
But if this caused any difficulties, we can definitely have that as a requirement in README

gdams (approved):

sxa (approved):
LGTM - although mostly for the link syntax fixes :-)

PR #492, opened 2022-06-26
Build and test with Java 17

luozhenyu:
# Changes
1. Upgrade gradle from 6.9 to 7.4.2
2. Build and test with Java 17 instead of Java 11


karianna (approved):
LGTM

PR #496, opened 2022-06-29
Update supported suse distributions

luozhenyu:
1. openSUSE Leap 15.1/15.2 have reached their end of life and should be removed. openSUSE Leap 15.4 is available.
    * https://en.opensuse.org/Lifetime


karianna (approved):
lgtm

PR #502, opened 2022-07-26
Update installers for jdk11 to jdk-11.0.16+8

sophia-guo:
Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

PR #506, opened 2022-07-28
remove impish from linux matrix

gdams:
Impish went EOL on 14th July (https://ubuntu.com/about/release-cycle)

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -172,7 +172,7 @@ def uploadDebArtifacts(buildArch) {
         "s390x" : "s390x"
     ]
     def distro_list = ""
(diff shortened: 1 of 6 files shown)


tellison (approved):

tellison:
Part of https://github.com/adoptium/installer/pull/502

zdtsw (approved):

sxa (approved):
Alternative would be to update /etc/apt/sources.list to replace `*.ubuntu.com` with `old-releases.ubuntu.com` which will allow the repositories to be available again, but sine it's EOL (and not an LTS) I'm also happy with this approach.

PR #509, opened 2022-08-01
fix: gpg sign issue on RPM

zdtsw:
The bug was introduced when move the logic from Dockerfile to build.sh
which double %% is for syntax in Dockerfile so it need to be removed one in the build.sh

Ref: https://github.com/adoptium/adoptium/issues/152#issuecomment-1200800650


sxa (approved):
LGTM although it might be good to have `echo Signing RPM files` as a debugging aid for this in the future.
Will need https://github.com/adoptium/installer/pull/510 merged afterwards so we can republish
For future reference, the initial PR that caused the break that this is a solution for was https://github.com/adoptium/installer/pull/474

sxa:
Approval to ship during lockdown cycle please? (Committers can 👍🏻 for approval)

gdams (approved):

Issue #473, opened 2022-05-20
please increase version information on releasing new debian packages


harridu:
Apparently the Debian packages for Temurin 7, 11 and 17 have been changed a few days ago without increasing the version information (both package file name and internal version number visible in apt-cache). This leads to problems with reproducibility. Binaries that ought to be identical have different MD5SUMs. It makes Temurin literally unusable for any software escrow service.

karianna:
@gdams - I think you had a work around for this? (permanent fix in next release cycle)

harridu:
No

sxa:
@gdams Can you confirm what the permanent fix for this will be going forward so we can ensure it is documented in the release guide please?

gdams:
Closing as we have a permanent fix now. The bot will no longer be able to forcefully overwrite existing packages

gdams:
Closing as we have a permanent fix now. The bot will no longer be able to forcefully overwrite existing packages

gdams:
it shouldn't be possible with the token we use

PR #510, opened 2022-08-01
Bump RPM spec files to version 2 in order to allow rebuild for signing

sxa:
RPMs for 11.0.16+8 and 17.0.4+8 were published without being signed.
That issue is fixed by https://github.com/adoptium/installer/pull/509 and causes this problem:
```
Package temurin-17-jdk-17.0.4.0.0.8-1.aarch64.rpm is not signed
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: GPG check FAILED
```

This bumps the spec version so that we can republish in a way that means the the new fixed versions will be picked up by default.

### DIFF
--- a/linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
+++ b/linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
@@ -5,7 +5,7 @@
 #  $ rpmdev-vercmp 11.0.13.0.1___7 11.0.13.0.0+8
 #  11.0.13.0.0___8 == 11.0.13.0.0+8
 %global spec_version 11.0.16.0.0.8
-%global spec_release 1
+%global spec_release 2
 %global priority 1111
 
 %global source_url_base https://github.com/adoptium/temurin11-binaries/releases/download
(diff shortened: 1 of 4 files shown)


gdams (approved):
Change looks fine but I'd suggest that you add a comment reminding people to reset this to `1` for the next round of releases

gdams (changes requested):
Just remembered that you'll need to update the changelog too

karianna (approved):
Approved (but please do address George's comments as well)

sxa:
@karianna @gdams Which changelog is appliable to the RPMs?

This is another example of why we need to get https://github.com/adoptium/installer/issues/473 closed off with the appropriate information for al distributions in it.

sxa:
> @karianna @gdams Which changelog is appliable to the RPMs?

OK Now understood to be the integral one in the spec files, but what are the implications on the version number - I presume the numeric suffix on `adopt0` is the build number rather than the spec version. What is the recommended format of the changelog for an installer update?

andrew-m-leonard (approved):

gdams (approved):

PR #521, opened 2022-08-23
Update jdk11 to jdk11.0.16.1+1

sophia-guo:


sophia-guo:
Not sure if we are still under lockdown period. 

Requesting approval to merge during the lockdown period. Please thumbs up the comment to approve

smlambert (approved):
We are no longer locked down, and for the PRs relating to release (and mentioned in the Release Checklist), they should be considered exempt from the lockdown, IMHO.

PR #542, opened 2022-10-17
linux: remove unnecessary update commands

gdams:
the yum/dnf update commands take a long time and slow down the CI unnecessarily.

### DIFF
--- a/linux/jdk/redhat/src/packageTest/java/packaging/DnfOperationsTest.java
+++ b/linux/jdk/redhat/src/packageTest/java/packaging/DnfOperationsTest.java
@@ -50,9 +50,6 @@ class DnfOperationsTest {
 
 			Container.ExecResult result;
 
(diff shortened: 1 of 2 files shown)


karianna (commented):
Can we replace with something more lightweight as opposed to removing entirely?

gdams:
> Can we replace with something more lightweight as opposed to removing entirely?

It's simply not required for the test, there is no need for us to verify that dnf/yum update works on a Linux distro (it's safe to say that any problems with that would be a problem for the upstream distro to resolve)

karianna (approved):
LGTM

gdams:
Testing codefreeze

gdams:
Good bot

gdams:
Going to request merge exception for this PR as it will significantly speed up the Linux installer release process. Risk is very low.

smlambert (approved):

Issue #549, opened 2022-11-04
remove zypper update in the test: we do not test package management function (false alarm) and it takes longer time to finish case


zdtsw:
similar to https://github.com/adoptium/installer/pull/542
and spot failed case because of this
- change is for jdk19 but jdk17 suse failed https://github.com/adoptium/installer/runs/9285518023#r9
- change is for jdk17 but jdk19_suse failed https://github.com/adoptium/installer/pull/534/checks?check_run_id=9285370789
- change for jdk8 failed on jdk17 19 suse https://github.com/adoptium/installer/actions/runs/3392236390/jobs/5638203391 

<p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(36, 41, 47); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong style="box-sizing: border-box; font-weight: 600;">3</strong><span> </span>tests were completed in<span> </span><strong style="box-sizing: border-box; font-weight: 600;">126s</strong><span> </span>with<span> </span><strong style="box-sizing: border-box; font-weight: 600;">2</strong><span> </span>passed,<span> </span><strong style="box-sizing: border-box; font-weight: 600;">1</strong><span> </span>failed and<span> </span><strong style="box-sizing: border-box; font-weight: 600;">0</strong><span> </span>skipped.</p>

Test suite | Passed | Failed | Skipped | Time
-- | -- | -- | -- | --
packaging.ZypperOperationsTest | 2 | 1 |   | 126s

[...]

PR #551, opened 2022-11-04
fix: wrong function name and target name

zdtsw:
from https://github.com/adoptium/installer/commit/c2d0c59f57a7dc4b269eaca70f4adf0979630512
name was change from uploadRPMArtifacts to uploadRpmArtifacts

it should be ended in arm7hl not arm7l
which showed the same for Sep jdk19 release

remove zypper update test which fails very often and slow down release
https://github.com/adoptium/installer/issues/549

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -282,7 +282,7 @@ def uploadDebArtifacts(buildArch) {
     )
 }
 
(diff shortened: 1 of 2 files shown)


gdams (approved):

karianna (approved):
LGTM

PR #552, opened 2022-11-05
fix: only have one arch defined in Deb, make sure the node label correct

zdtsw:
- old solution loops thro each char in string which makes no label on the node available


karianna (approved):
LGTM

gdams (approved):

PR #569, opened 2022-12-27
get rid of %%pretrans since nothing uses it

kajtzu:
I noticed this while attempting to determine why osbuild-composer bombs out because temurin-17-jdk dependencies seem to not be satisfied when osbuild's org.osbuild.rpm phase attempts to install packages in an empty build root.

The temurin jdk spec files have an empty %%pretrans section, meaning there isn't anything to be run in it, but that doesn't mean rpm would not try to process that section when starting the install transaction ;( This is because in an empty build root there is absolutely nothing yet installed.

In any case, it is severely discouraged to use %%pretrans at all and most up-to-date rpm documentation seem to prohibit its use with a shell interpreter.

Please see https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#pretrans and https://github.com/osbuild/osbuild-composer/issues/3197 and and https://bugzilla.redhat.com/show_bug.cgi?id=1683365 and and https://rpm-software-management.github.io/rpm/manual/lua

### DIFF
--- a/linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
+++ b/linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
@@ -5,7 +5,7 @@
 #  $ rpmdev-vercmp 11.0.13.0.1___7 11.0.13.0.0+8
 #  11.0.13.0.0___8 == 11.0.13.0.0+8
 %global spec_version 11.0.17.0.0.8
-%global spec_release 1
+%global spec_release 2
 %global priority 1111
 
 %global source_url_base https://github.com/adoptium/temurin11-binaries/releases/download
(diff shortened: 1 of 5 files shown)


karianna (approved):
LGTM

gdams (approved):

gdams:
Seems sane to me, we'll have to rebuild the RPM's and push them back up

kajtzu:
asking to understand - will this require something from me or will one of you/someone else/something else do the actual merge at some point in the future? kthx

gdams:
> asking to understand - will this require something from me or will one of you/someone else/something else do the actual merge at some point in the future? kthx

Yeah that's something we'll need to do when we merge. I'm going to give people a few more days to comment on this PR and then we'll merge it. Thanks!

PR #603, opened 2023-02-04
switch windows installer background to transparent PNG

gdams:
Currently, the installer is not accessible when running in high contrast mode as the white background BMP image clashes with the white text. The WIX toolset has supported PNG backgrounds for some time so it makes sense to change.

Also updated the binaries in the GitHub workflow and uploaded the PSD files for future editing.

Before:
Screenshot 2023-02-04 at 10 31 21

After:
Screenshot 2023-02-04 at 10 31 02


zdtsw (approved):

smlambert (approved):
👍

Issue #582, opened 2023-01-04
Incorrect Repository URL for OracleLinux 7 Server


rexxin-de:
### Please provide a brief summary of the bug

Following the instructions from https://adoptium.net/installation/linux/ does not work for OracleLinux 7 because yum/dnf variable "releasever" differs from the https://packages.adoptium.net/ path.

### Please provide steps to reproduce where possible

The following instructions from https://adoptium.net/installation/linux/ were performed:

```
# Uncomment and change the distribution name if you are not using CentOS/RHEL/Fedora
DISTRIBUTION_NAME=oraclelinux

cat <<EOF > /etc/yum.repos.d/adoptium.repo
[Adoptium]
name=Adoptium
baseurl=https://packages.adoptium.net/artifactory/rpm/${DISTRIBUTION_NAME:-$(. /etc/os-release; echo $ID)}/\$releasever/\$basearch
enabled=1
gpgcheck=1
gpgkey=https://packages.adoptium.net/artifactory/api/gpg/key/public
EOF

yum install temurin-8-jdk
```

### Expected Results

Package temurin-8-jdk should be installed via yum.

### Actual Results

yum exits with error:

```
# yum install temurin-8-jdk
Loaded plugins: ulninfo
https://packages.adoptium.net/artifactory/rpm/oraclelinux/7Server/x86_64/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
Trying other mirror.

 One of the configured repositories failed (Adoptium),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:
[...]

failure: repodata/repomd.xml from Adoptium: [Errno 256] No more mirrors to try.
https://packages.adoptium.net/artifactory/rpm/oraclelinux/7Server/x86_64/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
```

### What Java Version are you using?

None

### What is your operating system and platform?

OracleLinux 7 Server x86_64

### How did you install Java?

rpm package through Adoptium repository + yum

PR #612, opened 2023-02-13
fix: path in artifactory for oraclelinux

zdtsw:
Ref: https://github.com/adoptium/installer/issues/582

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -349,8 +349,8 @@ def uploadRpmArtifacts(String DISTRO, String rpmArch) {
             'rpm/fedora/35',
             'rpm/fedora/36',
             'rpm/fedora/37',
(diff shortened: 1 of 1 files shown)


gdams:
Noting that this is a breaking change, probably needs the PMC to talk through this one

zdtsw:
> Noting that this is a breaking change, probably needs the PMC to talk through this one

convert to "draft" till PMC's decision on #582

zdtsw:
any conclusion for this ?

karianna:
Added to agenda for next meeting

smlambert:
PMC call March 8th, since this PR would break people, suggest to not go this route / close PR.

Preferred approach is to update the documentation that is not correct (as reported in https://github.com/adoptium/installer/issues/582).

zdtsw:
close PR as it is not needed

PR #618, opened 2023-02-22
Fix: Update Spec Numbers For Signed RPM Re-Release

steelhead31:
Update Spec Numbers For Signed RPM Re-Release, this is only required for RHEL / SUSE, as both APK and DEB packages do not get signed.

### DIFF
--- a/linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
+++ b/linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
@@ -5,7 +5,7 @@
 #  $ rpmdev-vercmp 11.0.13.0.1___7 11.0.13.0.0+8
 #  11.0.13.0.0___8 == 11.0.13.0.0+8
 %global spec_version 11.0.18.0.0.10
-%global spec_release 2
+%global spec_release 3
 %global priority 1111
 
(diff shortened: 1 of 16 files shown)


gdams (approved):

linux/jdk/redhat/src/main/packaging/temurin/11/temurin-11-jdk.spec
```diff
@@ -262,6 +262,8 @@ fi
 /usr/lib/tmpfiles.d/%{name}.conf
 
 %changelog
+* Wed Feb 22 2023 Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org> 11.0.18.0.0.10.adopt0
```
jiekang, line 265: The changelog version should have the full release string which would include the `spec_release`, for example `11.0.18.0.0.8-1`. I'm not sure why there is adopt0 appended to the changelog entries, it doesn't seem to be used anywhere else.

steelhead31: Thanks @jie,  I'll amend the version strings, but leave adopt0 in place for the moment!

jiekang (approved):

zdtsw (approved):

PR #636, opened 2023-03-20
linux installer: generate matrix dynamically

gdams:
Currently whenever Linux changes are made we run all GH actions testing regardless of the changed files. This PR makes the action slightly more "intelligent" and will only trigger build jobs on PR's that change certain files.

I'd like to get this in before we publish the new installers as it will be the best time to test them


andrew-m-leonard (approved):
lgtm

PR #646, opened 2023-04-03
Added abilitiy to build RPM and DEB files locally

jmjaffe37:
Title


linux/jdk/debian/build.gradle
```diff
@@ -109,7 +127,12 @@ task checkJdkDebian(type: Test) {
 	def product = getProduct()
 	def productVersion = getProductVersion()
 
-	environment "PACKAGE", "${product}-${productVersion}-jdk"
+	// If product is microsoft set it as msopenjdk
```
karianna, line 130: I think this block is actually removing the suffix as $product would work?

linux/jdk/redhat/build.gradle
```diff
@@ -57,18 +57,40 @@ task packageJdkRedHat {
 	def gpgKey = getGPGKey()
 	def arch = getArch()
 
+	def compressed_jdk_distribution_dir = getCompressedJdkDistributionDir()
+
+
```
karianna, line 62:
```suggestion
```

linux/jdk/redhat/build.gradle
```diff
@@ -115,7 +137,13 @@ task checkJdkRedHat(type: Test) {
 	def gpgKey = getGPGKey()
 	def arch = getArch()
 
-	environment "PACKAGE", "$product-$productVersion-jdk"
+	// If product is microsoft set it as msopenjdk
+	if (product == "microsoft") {
```
karianna, line 141: Same comment as previously

karianna (changes requested):

gdams:
Same comment as the other PR, please can you add the same support to the `/linux/jre` directory too

linux/jdk/debian/build.gradle
jmjaffe37: I'm not entirely sure what you mean here, but I also included these lines in this PR by mistake. They actually belong in [PR 648](https://github.com/adoptium/installer/pull/648), which is why that PR failed on first check. I will move these lines to that PR and try again.

Also, it this clears up confusion: Micrsoft's `product`="Microsoft" and `product_version`="17" | "11". Without these lines, the package would be named `microsoft-17-jdk`, but we need our package to be named `msopenjdk-17`|`msopenjdk-11`.

linux/jdk/redhat/build.gradle
jmjaffe37: Answered above

karianna:
> Note: If we change [this line](https://github.com/adoptium/installer/pull/646/files#diff-e6b9dd1bc6d6f4e60e402932a3306783e062cc8683f364a7fc94bf238cacbab9L22) from: `rpmbuild --nodeps -bs "$spec"; # build src.rpm` to `rpmbuild --define "local_build ${buildLocalFlag}" \` ` --nodeps -bs "$spec"; # build src.rpm`
> 
> (And pretty much the [same change here](https://github.com/adoptium/installer/pull/646/files#diff-e6b9dd1bc6d6f4e60e402932a3306783e062cc8683f364a7fc94bf238cacbab9R47))
> 
> Then this will simplify the logic in the spec files for determining whether or not a vendor is building locally or not. Currently, my workaround can be seen in the [Microsoft installers PR](https://github.com/adoptium/installer/pull/648/files#diff-ef409bc76655b45f4198ab45113ea9adea99587d4fb0062f4f73092146648720R16), but it is a little ugly.
> 
> This line (the definition for the `local_build` variable) would simplify to be written the same way as the definition of `override_arch_ ` (currently on line 19 of the same file) if the `--define` change is ok with everyone. I believe that defining this variable (which I arbitrarily named), should not mess with anyone's builds since it is up to the specfile to add logic to use this variable or not, but wanted to ask what others think before making this change

That seems sane to me, I'm not sure if we would need to change any calling scripts in the Adoptium pipelines though.  Let's try the change and see if it passes the GH checks and then figure out if there is a caller script in Jenkins that also would need updating.

jmjaffe37:
> > Note: If we change [this line](https://github.com/adoptium/installer/pull/646/files#diff-e6b9dd1bc6d6f4e60e402932a3306783e062cc8683f364a7fc94bf238cacbab9L22) from: `rpmbuild --nodeps -bs "$spec"; # build src.rpm` to `rpmbuild --define "local_build ${buildLocalFlag}" \` ` --nodeps -bs "$spec"; # build src.rpm`
> > (And pretty much the [same change here](https://github.com/adoptium/installer/pull/646/files#diff-e6b9dd1bc6d6f4e60e402932a3306783e062cc8683f364a7fc94bf238cacbab9R47))
> > Then this will simplify the logic in the spec files for determining whether or not a vendor is building locally or not. Currently, my workaround can be seen in the [Microsoft installers PR](https://github.com/adoptium/installer/pull/648/files#diff-ef409bc76655b45f4198ab45113ea9adea99587d4fb0062f4f73092146648720R16), but it is a little ugly.
> > This line (the definition for the `local_build` variable) would simplify to be written the same way as the definition of `override_arch_ ` (currently on line 19 of the same file) if the `--define` change is ok with everyone. I believe that defining this variable (which I arbitrarily named), should not mess with anyone's builds since it is up to the specfile to add logic to use this variable or not, but wanted to ask what others think before making this change
> 
> That seems sane to me, I'm not sure if we would need to change any calling scripts in the Adoptium pipelines though. Let's try the change and see if it passes the GH checks and then figure out if there is a caller script in Jenkins that also would need updating.

I added the change

jmjaffe37:
> Same comment as the other PR, please can you add the same support to the `/linux/jre` directory too

Feature added :)

gdams (approved):

karianna (approved):
LGTM - but I wonder if the logic to build locally can be captured more cleanly inside namedd functions / blocks, something like:

function buildLocal() {...}

and checks like

if (isLocal())

jmjaffe37:
> LGTM - but I wonder if the logic to build locally can be captured more cleanly inside namedd functions / blocks, something like:
> 
> function buildLocal() {...}
> 
> and checks like
> 
> if (isLocal())

I'm not fully sure what you mean by this, but maybe I can attempt to explain my original approach here and see if that's what the team might prefer (also not sure if my explanation will make sense in text form here).

Originally in `linux/build.gralde`, I had a function `getLocalBuildStatus()` which returns 'true' if an input directory was provided, otherwise false.

Then, in `linux/jdk/<debian | redhat>/build.gradle`, in the `project.exec` command: I passed through an additional environment variable `buildLocalFlag` which was used to determine whether or not to build locally. I decided against this since I could get it to work without reliance across these extra files.

karianna (approved):

PR #647, opened 2023-04-03
Fixed infrequent 'permission denied' bug

jmjaffe37:
Fixed infrequent 'permission denied' bug when reading/writing to output folder for rpm or deb builds


karianna (approved):

gdams (changes requested):
can you please apply the same changes in the `/linux/jre` config files too?

jmjaffe37:
> can you please apply the same changes in the `/linux/jre` config files too?

I added the requested change. I also added this to the entrypoint.sh files for suse (jdk and jre) in a separate commit. If this is undesired, I can remove it. The only file left of this kind without this fix is `linux\ca-certificatres\debian\src\main\packaging\entrypoint.sh` but I don't think it needs it. (I am also unsure if suse needs it, but I figured it might be better to get like files to conform)

gdams (approved):

PR #648, opened 2023-04-03
Added Microsoft's installer files for linux rpm and deb

jmjaffe37:
Let me know if I should break this PR in half (one for RPM, one for DEB)


linux/jdk/debian/src/main/packaging/microsoft/11/debian/control
```diff
@@ -0,0 +1,62 @@
...
+          java11-runtime-headless,
+          java2-runtime-headless,
+          java5-runtime-headless,
+          java6-runtime-headless,
+          java7-runtime-headless,
+          java8-runtime-headless,
+          java9-runtime-headless
+Description: Microsoft Build of OpenJDK 11
+ Microsoft Build of OpenJDK is an OpenJDK-based development environment to create
+ applications and components using the programming language Java.
```
karianna, line 62:
```suggestion
 applications and components using the Java programming language.
```

linux/jdk/debian/src/main/packaging/microsoft/17/debian/compat
```diff
@@ -0,0 +1 @@
+11
```
karianna, line 1: I assume this is a debian compatibility value and therefore 11 makes sense

linux/jdk/debian/src/main/packaging/microsoft/11/debian/control
karianna: If accepted then apply this change for 17 also.

karianna (changes requested):

linux/jdk/debian/src/main/packaging/microsoft/17/debian/compat
gdams: yup it is

linux/jdk/debian/src/main/packaging/microsoft/11/debian/control
jmjaffe37: Accepted 👍

karianna:
> Another note: it seems like the RPM package checks are failing because they are attempting to build and test on `ppc64le` architecture (see [this line in the logs](https://github.com/adoptium/installer/actions/runs/4611073117/jobs/8150312199?pr=648#step:4:1949)). Is there a way to build and test on a different architecture (like x86_64), or should I add support for `ppc64le`?

We should change the tests so that it doesn't try to build for ppc64le for the Microsoft builds

gdams:
> > Another note: it seems like the RPM package checks are failing because they are attempting to build and test on `ppc64le` architecture (see [this line in the logs](https://github.com/adoptium/installer/actions/runs/4611073117/jobs/8150312199?pr=648#step:4:1949)). Is there a way to build and test on a different architecture (like x86_64), or should I add support for `ppc64le`?
> 
> 
> 
> We should change the tests so that it doesn't try to build for ppc64le for the Microsoft builds

@jmjaffe37 I'll look at the code for this

linux/jdk/debian/src/main/packaging/microsoft/11/debian/control
```diff
@@ -0,0 +1,62 @@
+Source: msopenjdk-11
+Section: java
+Priority: optional
+Maintainer: Microsoft Package Maintainers <openjdk@microsoft.com>
+Build-Depends: debhelper (>= 11), lsb-release
+
+Package: msopenjdk-11
+Architecture: amd64 arm64
+Depends: adoptium-ca-certificates,
```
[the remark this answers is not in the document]
jmjaffe37: I agree, but I would need more guidance to make these changes appropriately

gdams:
> > > Another note: it seems like the RPM package checks are failing because they are attempting to build and test on `ppc64le` architecture (see [this line in the logs](https://github.com/adoptium/installer/actions/runs/4611073117/jobs/8150312199?pr=648#step:4:1949)). Is there a way to build and test on a different architecture (like x86_64), or should I add support for `ppc64le`?
> > 
> > 
> > We should change the tests so that it doesn't try to build for ppc64le for the Microsoft builds
> 
> @jmjaffe37 I'll look at the code for this

@jmjaffe37 I've pushed a fix to this branch but it's pretty fragile, I've raised an alternative fix here: https://github.com/adoptium/installer/pull/650 which would allow us to define the supported architectures for each specfile rather than making endless changes to the already growing build.sh file

karianna (approved):
LGTM, we can solved the certs issue internally and then upstream a change if required.

gdams (approved):
PMC was happy with this so LGTM

PR #650, opened 2023-04-05
feat: switch spec files to define ExclusiveArch

gdams:
This allows us to automatically parse the list of arches to build from the specfile rather than writing if statements


zdtsw (approved):

tellison (approved):
Good move

PR #669, opened 2023-05-05
Updated deb and rpm config files for microsoft vendor

jmjaffe37:
These changes reflect what we now use internally

### DIFF
--- a/linux/jdk/debian/src/main/packaging/microsoft/11/debian/changelog
+++ b/linux/jdk/debian/src/main/packaging/microsoft/11/debian/changelog
@@ -1,11 +1,11 @@
-msopenjdk-11 (11.0.19.0.0+7) STABLE; urgency=medium
+msopenjdk-11 (11.0.19-1) STABLE; urgency=medium
 
-  * Microsoft 11.0.19.0.0+7 release.
+  * Microsoft 11.0.19-1 release.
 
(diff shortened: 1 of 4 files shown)


linux/jdk/debian/src/main/packaging/microsoft/11/debian/changelog
```diff
@@ -1,11 +1,11 @@
-msopenjdk-11 (11.0.19.0.0+7) STABLE; urgency=medium
+msopenjdk-11 (11.0.19-1) STABLE; urgency=medium
```
karianna, line 1: Remind me again why the `-1` vs the Adoptium versioning scheme?

karianna (approved):

linux/jdk/debian/src/main/packaging/microsoft/11/debian/changelog
jmjaffe37: This reflects the versioning that we have used since before we updated to the new linux installers creation

karianna: OK, worth having a chat to Bruno and see if he wants to keep that versioning scheme or move to Adoptium's scheme

karianna: Keeping this naming scheme for MSFT's Mariner distro

gdams (approved):

Issue #670, opened 2023-05-10
Debian Bookworm repository lacks packages


Montandalar:
### Please provide a brief summary of the bug

I have added the adoptium package mirror for Debian bookworm to my apt sources and run apt update, but it lacks packages I want.

 Now when I search for packages I notice there is no temurin-8-jdk and no adoptium-ca-certificates. I have confirmed this by navigating my web browser to https://packages.adoptium.net/ui/native/deb/dists/bookworm/main/binary-amd64/Packages and noting the lack of these packages, whereas under the Bullseye `Package` file both of those packages are present.

The ca-certificates package is needed as a dependency for all the other packages, and I was after the JDK 8 not just the JRE. I will have to use the .tar.gz distribution instead now, but would prefer to use the apt mirror.

### Please provide steps to reproduce where possible

```
sudo echo "deb https://packages.adoptium.net/artifactory/deb bookworm main" > /etc/apt/sources.list.d/adoptium.list
sudo apt-get update
sudo apt install temurin-8-jdk
```

### Expected Results

Installation of that package

### Actual Results

```
$ sudo apt install temurin-8-jdk
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package temurin-8-jdk
$ sudo apt install temurin-8-jre
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 temurin-8-jre : Depends: adoptium-ca-certificates but it is not installable
E: Unable to correct problems, you have held broken packages.
```

### What Java Version are you using?

irrelevant

jerboaa:
Could be related to https://github.com/adoptium/installer/issues/668

sxa:
Thanks for raising this - a temporary workaround is to use `bullseye` in the `/etc/apt/sources.list.d/adoptium.list` instead of `bookworm`.

syhe:
I only now discovered this issue. I have already reported the missing package at adoptium/adoptium-support/issues/816 – should that issue be closed in favor of this one?

Montandalar:
Now that Debian Bookworm has reached a stable release, delivering packages for it can probably be considered a much bigger priority, right?

PR #591, opened 2023-01-21
Feature/430 Add Temurin JRE Installers

ascheman:
Please **do not squash commits** as discussed in [The Adoptium community call of 2023, Jan 18th](https://docs.google.com/document/d/1oaU8GA_3F8DDzjeZujY10fiyRS-vKdF5rE0eAhAascY/edit#) - all commits are curated and make a substantial change on their own.
Hence, they may be  rolled back (reverted) separately.

This PR closes #430, #584, #585, #590 and contributes to adoptium/adoptium#203, #575

PR #683, opened 2023-06-13
Add Debian Bookworm support to ca-certificates package

sxa:
Fixes https://github.com/adoptium/adoptium-support/issues/816 and is the final step in resolving https://github.com/adoptium/installer/issues/670 (Note: Most of the support was added in https://github.com/adoptium/installer/pull/591 but the ca-certificates package does not get pushed to new directories unless a version change it made, which is what this PR will do.


gdams (approved):

sxa:
(Pushed update to adjust the test which was failing as it was still looking for 1.0.1-1)

andrew-m-leonard (approved):

PR #706, opened 2023-08-02
Fix Mapping Of arm32 parameter in jenkins to value required

steelhead31:
Fix the mapping Of arm32 parameter in jenkins (armv7l) to value required in artifactory (armv7hf)


sxa (approved):
Reverses the change from https://github.com/adoptium/installer/commit/920af118f850a8f1318c61606b546e70607c4b0c - SGTM

PR #708, opened 2023-08-02
Add additional armv7hl parameter as workaround for arm32, fix arm32 jre dates

steelhead31:
Add additional armv7hl parameter as workaround for arm32, fix jre dates


sxa (approved):
Yuk :-) Although I'm not sure we need a SuSE verison on arm32

adamfarley (approved):
LGTM

PR #722, opened 2023-08-29
Create Process/Tool to Validate Upload Of Release Artifacts To Artifactory

Ndacyayisenga-droid:
cc @steelhead31

### DIFF
--- a/linux/scripts/artifacts.sh
+++ b/linux/scripts/artifacts.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+TYPE=("jdk" "jre" "src")
+VERS=("$@") # Accept the VERS array as arguments
+DIST=("apk" "deb" "rpm")
+DEBDISTS=("bookworm" "bullseye" "buster" "kinetic" "jammy" "focal" "bionic")
+RPMDISTS=("centos/7" "rocky/8" "rhel/7" "opensuse/15.3")
+
+if [ $# -eq 0 ]; then
+    echo "Please provide the VERS array as arguments"
+    exit 1
+fi
+
+for EACHTYPE in "${TYPE[@]}"; do
+    for EACHVERS in "${VERS[@]}"; do
+        for EACHDIST in "${DIST[@]}"; do
+            ## Limit APK To Only x86_64
+            if [ "$EACHDIST" = "apk" ]; then
+                ARCH=("x86_64")
+            else
+                ARCH=("x86_64" "armv7hl" "armv7l" "aarch64" "ppc64le" "s390x")
+            fi
+            for EACHARCH in "${ARCH[@]}"; do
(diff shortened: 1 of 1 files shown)


linux/scripts/artifacts.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+TYPE=("jdk" "jre" "src")
+VERS=("$@") # Accept the VERS array as arguments
+DIST=("apk" "deb" "rpm")
+DEBDISTS=("bookworm" "bullseye" "buster" "kinetic" "jammy" "focal" "bionic")
+RPMDISTS=("centos/7" "rocky/8" "rhel/7" "opensuse/15.3")
+
+if [ $# -eq 0 ]; then
```
hendrikebbers, line 9: Such a script should have a documentation that is printed to the Command-line. The documentation should contain the functionality of the script and possible parameters with documentation.

Example
```
usage="$(basename "$0") [-h] [-s n] -- program to calculate the answer to life, the universe and everything

where:
    -h  show this help text
    -s  set the seed value (default: 42)"

seed=42
while getopts ':hs:' option; do
  case "$option" in
    h) echo "$usage"
       exit
       ;;
    s) seed=$OPTARG
       ;;
    :) printf "missing argument for -%s\n" "$OPTARG" >&2
       echo "$usage" >&2
       exit 1
       ;;
   \?) printf "illegal option: -%s\n" "$OPTARG" >&2
       echo "$usage" >&2
       exit 1
       ;;
  esac
done
shift $((OPTIND - 1))
```
Copied from https://stackoverflow.com/questions/5474732/how-can-i-add-a-help-method-to-a-shell-script

Ndacyayisenga-droid: > Such a script should have a documentation that is printed to the Command-line. The documentation should contain the functionality of the script and possible parameters with documentation.
> 
> Example
> 
> ```
> usage="$(basename "$0") [-h] [-s n] -- program to calculate the answer to life, the universe and everything
> 
> where:
>     -h  show this help text
>     -s  set the seed value (default: 42)"
> 
> seed=42
> while getopts ':hs:' option; do
>   case "$option" in
>     h) echo "$usage"
>        exit
>        ;;
>     s) seed=$OPTARG
>        ;;
>     :) printf "missing argument for -%s\n" "$OPTARG" >&2
>        echo "$usage" >&2
>        exit 1
>        ;;
>    \?) printf "illegal option: -%s\n" "$OPTARG" >&2
>        echo "$usage" >&2
>        exit 1
>        ;;
>   esac
> done
> shift $((OPTIND - 1))
> ```
> 
> Copied from https://stackoverflow.com/questions/5474732/how-can-i-add-a-help-method-to-a-shell-script

Thannks for the pointer @hendrikebbers. Actually there is still more work to this script. I think we can consider this as well

PR #723, opened 2023-08-31
Installer updates for JDK11 & JDK17 for August 2023 respin releases

steelhead31:
Update APK, deb and rpm spec files for release.

PR #742, opened 2023-09-27
Added jdk21 config files for msft linux deb+rpm installers

jmjaffe37:
These files were used internally to get our release out


d3r3kk (approved):

karianna:
@jmjaffe37 Looks like failing tests, is that expected?

jmjaffe37:
@karianna it looks like the failing test is checking to see if this package can be downloaded on ubuntu 22.10, but I don't think we support that release. I will try to find the source code of the test for more info

jmjaffe37:
It seems to be erroring on [this assertion](https://github.com/adoptium/installer/blob/61e4cf765d6202efbedf36f62bd74ecbeea171ea/linux/jdk/debian/src/packageTest/java/packaging/AptOperationsTest.java#L59). For only ubuntu kinetic, there is a non-zero exit code when doing `apt-get update` on ubuntu kinectic. The same issue can be seen for jdk11 and 17 in [this PR](https://github.com/adoptium/installer/pull/743)

jmjaffe37:
When I ran these tests locally, the error logs tell me that there is an execution error Caused by: `java.lang.OutOfMemoryError: Java heap space`

jmjaffe37:
This issue seems to affect all debian installer creation processes, see this GHA run from my fork using this branch: https://github.com/jmjaffe37/openjdk-installer/actions/runs/6331915451

jmjaffe37:
Also, here is a link to the last clean run of the PR checks: https://github.com/adoptium/installer/pull/723

There are only 5 closed PRs ahead of this one, so the hunt should be narrow. Even so, I glanced at those PRs and it feels like they should not have caused this error

jmjaffe37:
Looks like this is due to ubuntu Kinetic hitting EoL and shutting down servers this month. We have 2 options:

1. Removing ubuntu kinectic from list of ubuntu flavors in adoptium/installers
2. Following [this stackoverflow post](https://askubuntu.com/questions/1458489/ubuntu-22-10-kinetic-cant-update-archive-ubuntu-com-and-ignored-it) to get ubuntu:kinetic to continue working past EoL

karianna:
I would recommend 1 and send a note to the adoptium installer channel to that effect

jmjaffe37:
@karianna sorry for the typo in my previous comment (it is fixed now), I actually implemented a solution following option 1 in my PR

karianna:
looks like 21 fails but that's a race condition with your other PR I assume?

linux/jdk/debian/src/main/packaging/microsoft/21/debian/rules
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/make -f
+
+pkg_name = msopenjdk-21
+priority = 1911
```
Luigi96, line 4: Do we want to change the default priority of this package? I am not sure about this.
```suggestion
priority = 2111
```

karianna (approved):

linux/jdk/debian/src/main/packaging/microsoft/21/debian/rules
jmjaffe37: I'm ok with this change, although we may need to change it again when we see what the community agrees on

Luigi96 (approved):

Issue #744, opened 2023-09-27
All debian installer checks fail due to Ubuntu:Kinetic EoL


jmjaffe37:
The task checkJdkDebian consistently fails when building locally or performing GHA checks for a PR. This is dues to Ubuntu:Kinetic (22.10) reaching EoL in July 2023, and shutting off there servers in September 2023. As a result, Ubuntu:Kinetic will fail to run apt-get update.

This attached run shows that the issue is present in all debian isntaller checks: https://github.com/jmjaffe37/openjdk-installer/actions/runs/6331915451

jmjaffe37:
I created [this PR](https://github.com/adoptium/installer/pull/745) which addresses the issue

jmjaffe37:
Just merged two PRs whose installer checks demonstrate that the issue was resolved as well:
https://github.com/adoptium/installer/pull/742
and
https://github.com/adoptium/installer/pull/743

PR #745, opened 2023-09-27
removed ubuntu 22.10 support due to EoL

jmjaffe37:
This fixes the failures in the PR automated checks due to ubuntu 22.10's end of life

### DIFF
--- a/.github/workflows/cacert-publish.yml
+++ b/.github/workflows/cacert-publish.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Upload deb file to Artifactory
         if: steps.check-deb.outputs.file_exists == 'false'
         run: |
-          debVersionList=("bookworm" "bullseye" "buster" "kinetic" "jammy" "focal" "bionic")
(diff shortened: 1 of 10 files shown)


jmjaffe37:
This GHA run (from my branch on my fork) shows that it fixes the problem outlined in the [issue 774](https://github.com/adoptium/installer/issues/744):

https://github.com/jmjaffe37/openjdk-installer/actions/runs/6332563576

karianna (approved):

gdams (approved):

karianna:
https://github.com/adoptium/containers/issues/359 and https://github.com/adoptium/containers/issues/423 are related to this.  I think we need to start a policy doc on the website

tellison:
Agreed, although I wasn't thinking of a _policy_ (this change is fine), rather the actual description of what is in/out of support for Temurin on Debian - [as we do with other OSes](https://adoptium.net/supported-platforms/).

karianna:
Heh, had missed that, looks like we never stated we would support 22.10 (which isn't an LTS), I wonder if we mis stepped in adding it in the first place.

PR #747, opened 2023-10-02
Improve Resilience & Repeatability Of Linux Installers Packaging Process

### LINKED ISSUE #710: Improve installers upload process to check for existence of artifacts.
During the last 2 releases, there have been a number of instances where the upload of artifacts to artifactory has for one reason or another, failed after a partial upload. In this situation, the current installer process fails to run, as it fails as soon as it cant upload/update a file, and the process is ended, rather than skipping already completed artifacts, and just uploading the missing ones.

steelhead31:
Fixes #710 

Improve linux installer resilience and repeatability by checking for the presence of existing artifacts with the same name on the jfrog artifactory repository. Skip the upload whereby the files already exist.

Should a future packaging process fail during the upload, then the process should now be rerunnable, and not continue to fail when trying to upload a file that was uploaded during a previous run.


linux/Jenkinsfile
```diff
@@ -286,8 +286,59 @@ def uploadAlpineArtifacts(String buildArch) {
...
+          echo "Target Doesnt Exist - Upload Files"
+          jf 'rt u **/build/ospackage/temurin-*j*.apk apk/alpine/main/${BUILDARCH}/ --flat=true'
+        }
+
+      } catch (Exception e) {
+        error "Error While Checking URL ${Target}: ${e.message}"
+      }
+    }
+
+      for (SrcFile in SrcFiles) {
```
karianna, line 318: whitespace seems out here

linux/Jenkinsfile
```diff
@@ -338,10 +389,34 @@ def uploadDebArtifacts(String buildArch) {
...
+                  if ( ResponseCode == '200') {
+                    echo "Target Exists - Skipping"
+                  } else {
+                    echo "Target Doesnt Exist - Upload Files"
+                    jf 'rt u **/build/ospackage/temurin-*${BUILDARCH}.deb deb/pool/main/t/temurin-${VERSION}/ --target-props=${DISTROLIST}deb.component=main;deb.architecture=${BUILDARCH} --flat=true'
+                  }
+
+                } catch (Exception e) {
+                    error "Error While Checking URL ${Target}: ${e.message}"
+                  }
```
karianna, line 415: little bit ogf whitespace wonkiness here

linux/Jenkinsfile
```diff
@@ -338,10 +389,34 @@ def uploadDebArtifacts(String buildArch) {
         } else {
             env.BUILDARCH = debArchList[buildArch]
             env.DISTROLIST = distro_list
-            jf 'rt u **/build/ospackage/temurin-*${BUILDARCH}.deb deb/pool/main/t/temurin-${VERSION}/ --target-props=${DISTROLIST}deb.component=main;deb.architecture=${BUILDARCH} --flat=true'
+
+            def PackFiles = findFiles(glob: "**/build/ospackage/temurin-*${debArchList[buildArch]}.deb") // List All Packages To Upload
+            for (PackFile in PackFiles) {
+              def FileName = PackFile.name
```
karianna, line 395: this logic is similar to the APJ logic, worth exrtracting?

karianna (changes requested):

linux/Jenkinsfile
steelhead31: I did consider this, but given the globs, targets and upload commands are different for APK, DEB & RPM, the only element of abstraction would be the curl check, and not worth the added complexity. As Im hoping to further enhance this process in the near future, I think this is sufficient to solve the current problems we encounter at release time.

linux/Jenkinsfile
```diff
@@ -338,11 +387,34 @@ def uploadDebArtifacts(String buildArch) {
...
+              def Target = "https://adoptium.jfrog.io/artifactory/deb/pool/main/t/temurin-${VERSION}/${FileName}"
+
+              echo "File : ${PackFile}"
+              echo "Filename : ${FileName}"
+              echo "Target : ${Target}"
+
+              try {
+                def ResponseCode = sh(script: "curl -o /dev/null --silent --head --write-out '%{http_code}' ${Target}", returnStdout: true).trim()
```
gdams, line 401: we should use the JFrog CLI to check if a file exists rather than writing a custom curl script. Check out how I do this for the cacerts upload script:

https://github.com/adoptium/installer/blob/master/.github/workflows/cacert-publish.yml#L43-L52

```bash
if jf rt s "deb/pool/main/a/adoptium-ca-certificates/$(basename $FILE)" > /dev/null; then
  echo file_exists=true >> "$GITHUB_OUTPUT"
else
  echo file_exists=false >> "$GITHUB_OUTPUT"
fi
```

steelhead31: Ah, I had several attempts at this, but couldn't get it to work correctly, I'll try it using this method, see if I have any more success...

gdams: In Jenkins pipeline format it should look something like this:

```groovy
 jf 'rt s ${Target}'
```

steelhead31: Having spent the day, trying many different varieties, I think this is an issue or a lack of feature in the jfrog cli plugin for jenkins, the command line/shell driven normal version can be handled in the usual fashion, however the jenkins plugin, other than outputting test to the screen, does not provide anything that can be processed, so Im reverting this to the CURL based check , and will log issues with the provider, with the idea that this can be improved with a future update of the plugin.

linux/Jenkinsfile
```diff
@@ -286,10 +286,59 @@ def uploadAlpineArtifacts(String buildArch) {
...
+    for (PackFile in PackFiles) {
+
+      def FileName = PackFile.name
+      def Target = "https://adoptium.jfrog.io/artifactory/apk/alpine/main/${BUILDARCH}/${FileName}"
+
+      echo "File : ${PackFile}"
+      echo "Filename : ${FileName}"
+      echo "Target : ${Target}"
+
+      try {
```
gdams, line 302: this feels like a lot of duplicated code. Can you split it into a function and reuse it?

linux/Jenkinsfile
steelhead31: Reusable function added.

linux/Jenkinsfile
```diff
@@ -286,12 +325,23 @@ def uploadAlpineArtifacts(String buildArch) {
     // currently only support x64 as buildArch
     // set BUILDARCH environment variable for ${buildArch}
     env.BUILDARCH = buildArch
-    jf 'rt u **/build/ospackage/temurin-*j*.apk apk/alpine/main/${BUILDARCH}/ --flat=true'
-    jf 'rt u **/build/ospackage/temurin-*src*.apk apk/alpine/main/${BUILDARCH}/ --flat=true'
+
```
karianna, line 328:
```suggestion
```

linux/Jenkinsfile
```diff
@@ -286,12 +325,23 @@ def uploadAlpineArtifacts(String buildArch) {
...
+    def AllFiles = PackFiles + SrcFiles
+
+    for (PackFile in AllFiles) {
+      def FileName = PackFile.name
+      def Target = "https://adoptium.jfrog.io/artifactory/apk/alpine/main/${BUILDARCH}/${FileName}"
+      CheckAndUpload(Target, Distro, buildArch, '', '', '', '', '' )
+    }
     // unset BUILDARCH environment variable
-}
+  }
```
karianna, line 341:
```suggestion
}
```

linux/Jenkinsfile
```diff
@@ -338,13 +388,21 @@ def uploadDebArtifacts(String buildArch) {
...
+              def Target = "https://adoptium.jfrog.io/artifactory/deb/pool/main/t/temurin-${VERSION}/${FileName}"
+
+              CheckAndUpload(Target, Distro, BUILDARCH, VERSION, DISTROLIST, '', '', '' )
+            }
+          }
+          break
         }
-    }
-}
+      }
```
karianna, line 402:
```suggestion
}
```

linux/Jenkinsfile
```diff
@@ -408,10 +466,18 @@ def uploadRpmArtifacts(String DISTRO, String rpmArch) {
...
+                      def Target = "https://adoptium.jfrog.io/artifactory/${PACKAGEDIR}/${KEY}/Packages/${FileName}"
+
+                      CheckAndUpload(Target, Distro, rpmArchList.each, '', '', VALUE , PACKAGEDIR, KEY )
+                    }
                     break
                 }
             }
         }
-    }
+      }
```
karianna, line 482:
```suggestion
  }
```

karianna (approved):

gdams (approved):

PR #757, opened 2023-10-24
jdk-21 Oct CPU installers

sophia-guo:
Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

PR #763, opened 2023-10-26
Remove s390x, arm32 for jdk21 debian from jenkins job if arch=all

sophia-guo:


gdams (approved):

sxa:
I'm a little reluctant to approve this given s390x will be good for the next release, and hopefully we can look at pushing arm32 forward.
Since we're able to complete this for the 21.0.1 release without using `all` I would suggest we do that for this release and avoid having to remember to back these changes out and accidentally failing to ship the excluded platforms next time round.

sophia-guo:
Yes, we can complete the release without using all. However this change actually should be part of https://github.com/adoptium/installer/pull/757/files.

karianna:
@sophia-guo Do you want to close this PR then?

PR #768, opened 2023-10-30
Fix JDK17.0.9 Debian Version Numbers In The Changelog

steelhead31:
Fix JDK17.0.9 Debian Version Numbers In The Changelog

Updated from 17.0.9.0+9 to 17.0.9.0.0+9

### DIFF
--- a/linux/jdk/debian/src/main/packaging/temurin/17/debian/changelog
+++ b/linux/jdk/debian/src/main/packaging/temurin/17/debian/changelog
@@ -1,6 +1,6 @@
-temurin-17-jdk (17.0.9.0+9) STABLE; urgency=medium
+temurin-17-jdk (17.0.9.0.0+9) STABLE; urgency=medium
 
-  * Eclipse Temurin 17.0.9.0.+9 release.
+  * Eclipse Temurin 17.0.9.0.0+9 release.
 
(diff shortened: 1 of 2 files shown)


sxa (approved):
Obviously this came across as part of [this slack thread](https://adoptium.slack.com/archives/C9C272Z1T/p1698480502733769).

I'm ok to approve this but do we know if it will cause problems with people who already have the version which we had live? i.e. will this version with the extra .0 superceded the previous one? Since they're the same contents it shouldn't make too much of a difference but we would ideally want people to get the new one

steelhead31:
> Obviously this came across as part of [this slack thread](https://adoptium.slack.com/archives/C9C272Z1T/p1698480502733769).
> 
> I'm ok to approve this but do we know if it will cause problems with people who already have the version which we had live? i.e. will this version with the extra .0 superceded the previous one? Since they're the same contents it shouldn't make too much of a difference but we would ideally want people to get the new one

I've removed the erroneous ones from artifactory to prevent it impacting, it should still get updated on a future release, as 17.0.10 is later than either 17.0.9 version, Once the new artifacts are up, I'll post in the release channel advising the removal of the old, and replacement with the new, the identified failing update processes will work correctly with the new version.

smlambert (approved):

PR #771, opened 2023-11-01
Ensure version number is passed for Alpine & Debian installer builds

### LINKED ISSUE #764: jenkins apline and debian build failures when using arch=all for arch != x86_64
adoptium jenkins apline and debian build failures when using arch=all for arch != x86_64. The error message is like

```
+ ./gradlew packageJdkDebian --parallel -PPRODUCT=temurin -PPRODUCT_VERSION= -PARCH=armhf
Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* Where:
Build file '/home/jenkins/workspace/adoptium-packages-linux-pipeline_new/build.gradle' line: 40

* What went wrong:
A problem occurred evaluating project ':jdk:alpine'.
> For input string: ""
```

steelhead31:
Update jenkinsfile so that version number is correctly passed through to the Alpine and Debian gradle builds, when ARCH parameter is set to ALL, this has been tested in a non-destructive fashion, in my test harness.. https://ci.adoptium.net/job/scott-fryer-adoptium-packages-linux-pipeline_new/237/ , https://ci.adoptium.net/job/scott-fryer-adoptium-packages-linux-pipeline_new/236 , https://ci.adoptium.net/job/scott-fryer-adoptium-packages-linux-pipeline_new/234 & https://ci.adoptium.net/job/scott-fryer-adoptium-packages-linux-pipeline_new/230

Fixes #764

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -196,6 +196,7 @@ def CheckAndUpload(String Target, String Distro, String BuildArch, String Versio
 // function only handle debian as DISTRO
 def jenkinsStepDeb() {
     def temurinVersion = "${TYPE.toLowerCase()} ${VERSION} - ${ARCH}"
+    def VERSION = "${VERSION}"
     echo "Installer Job for Temurin ${temurinVersion} - Debian"
     //make sure this is an array not a string
     def debArchAllList = []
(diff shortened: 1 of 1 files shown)


steelhead31:
> I'm curious why VERSION is not accessible when doing parallel as it is an environment variable. Also interesting for debian and apline when running parallel with x86-64 and other platforms it consistently worked for x86-64. @steelhead31 do you have the idea?
> 
> I suspect that update these two lines https://github.com/adoptium/installer/blob/master/linux/Jenkinsfile#L324 and https://github.com/adoptium/installer/blob/master/linux/Jenkinsfile#L338 with -PPRODUCT_VERSION=${env.VERSION} probably also work.

I don't think it is an environment variable within the contexts used by Debian & Alpine, it sets correctly on the first iteration of a loop, but then is subsequently unavailable to future iterations. This fix pushes the PRODUCT VERSION into a local context to pass onwards. The RHEL & SUSE steps do use environment variables for this reason I believe.

karianna (approved):

gdams (approved):

Issue #600, opened 2023-01-31
Windows: Create link after upgrade or use fix path for each mayor version


LukasLeppich:
The path to the JDK changes after each upgrade. So you have to set all paths again after each upgrade and restart the applications.

Currently I use mklink on Windows to create a unique path for each Java version. But this step has to be done manually after each upgrade. This is very annoying, if the frustration becomes too much for some I fear that the installer will simply be removed and security will suffer. 

Can you please just use a fixed path for a Java major version or link from a fixed path to the new version after each upgrade?

I am a developer and use multiple Java versions at the same time, so it doesn't help me if the installer sets the environment variables. The new variables would only be active after a restart of the development environment and I would still have to manually select the correct Java version in each project.

gdams:
I think this might be a duplicate of https://github.com/adoptium/installer/issues/422

@douph1 is this something you might be able to take a look at implementing? It's come up as a pain-point for people a few times now.

PR #789, opened 2024-01-07
Changed INSTALLDIR to fixed path to fix #422 and #600

### LINKED ISSUE #422: Windows Installer: Provide reliable installation path
When installing any of the Adoptium versions on Windows a version specific (including patch level) path is used by default:
Screenshot 2022-01-23 at 12 08 28

If I understand e.g. #88 correctly then path variables and registry keys are updated on update. That's fine and a good beginning, however, insufficient in some use cases. For instance [Maven Toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html) are an optional, manual setup that tell Maven where to find major versions of JDKs that a user has installed. The paths are essentially hardcoded in the declaration file. So if I want my toolchains to always point to the latest JDK of a major release I'll manually have to update that after installation of the patch version.

I therefore propose to shorten the default installation path of Adoptium to the major version only, e.g. `C:\Program Files\Eclipse Adoptium\jdk-11.0.14.9-hotspot` becomes `C:\Program Files\Eclipse Adoptium\jdk-11-hotspot`. This would also align with macOS for example where the path is usually `/Library/Java/JavaVirtualMachines/temurin-11.jdk`.

On linux the `.deb` path is also major release only: 

```bash
$ update-java-alternatives -l
temurin-11-jdk-amd64           1111       /usr/lib/jvm/temurin-11-jdk-amd64
temurin-17-jdk-amd64           1161       /usr/lib/jvm/temurin-17-jdk-amd64
```

I guess this also relates to #325 in some way.

adipiciu:
By using only the Major version in the installdir path, the installation directory will not change during upgrades.

C:\Program Files\Eclipse Adoptium\jre-8-hotspot
C:\Program Files\Eclipse Adoptium\jdk-8-hotspot
C:\Program Files\Eclipse Adoptium\jre-11-hotspot
etc..

### DIFF
--- a/wix/Includes/OpenJDK.Variables.wxi.template
+++ b/wix/Includes/OpenJDK.Variables.wxi.template
@@ -27,7 +27,7 @@
     <?define FeatureMainDescription="!(loc.JREFeatureMainDescription)" ?>
 
     <?define ProductCategory="JRE" ?>
-    <?define AppFolder="jre-$(var.ProductVersion)-$(var.JVM)" ?>
+    <?define AppFolder="jre-$(var.ProductMajorVersion)-$(var.JVM)" ?>
   <?else?>
     <?define ProductName="!(loc.JDKProductName)" ?>
     <?define ProductNameWithVersion="!(loc.JDKProductName) $(var.ProductVersionString) $(var.Arch)" ?>
@@ -36,7 +36,7 @@
     <?define FeatureMainDescription="!(loc.JDKFeatureMainDescription)" ?>
 
     <?define ProductCategory="JDK" ?>
-    <?define AppFolder="jdk-$(var.ProductVersion)-$(var.JVM)" ?>
+    <?define AppFolder="jdk-$(var.ProductMajorVersion)-$(var.JVM)" ?>
   <?endif?>
 
   <!-- Registry key change with new numbering format since Java 9 -->


adipiciu:
Zulu, Liberica and Corretto are also using only the major version in the installdir path.

karianna:
what happens in the upgrade use case for existing installs?

adipiciu:
> what happens in the upgrade use case for existing installs?

The same thing as before, the installdir changes to the new one with a different version. But with this change, the installdir will remain fixed in the future.

karianna (approved):

sxa (commented):
We discussed this at the PMC today and generally agreed with this approach, however we should consider whether `jdk-xx-hotspot` is the right directory name for this or if we can select something better, so I'm opening this question to interested parties as to what the correct top level directory name should be.

sxa:
Comparing with Linux, which installs under /usr/lib/jvm and has directories named:

`temurin-17-jdk-amd64` (`.deb`)
`temurin-17-jdk` (`.rpm`)

On that basis it feels like a format change to `temurin-xx-jdk` etc. for windows would be appropriate - what do others think? Noting that while the Windows builds are using `hotspot` our build system typically has `temurin` as a value for `VARIANT` which is explicitly distint from `hotspot` so this would potentially be preferable.

adipiciu:
@sxa 'temurin-17-jdk' seems fine for me. But there is no temurin JVM variant. See #438

adipiciu:
Actually, Eclipse Adoptium is already in the path, so temurin is redundant. Right now, as mentioned here #2671, temurin = hotspot, but this was just a workaround. You should rely on the VENDOR variable to build a vanilla JDK, not on the JVM variant...
So the 'temurin-17-jdk' option is wrong.

sxa:
> as mentioned here https://github.com/adoptium/temurin-build/issues/2671, temurin = hotspot, but this was just a workaround

Yep - it was me that implemented that change - VARIANT was just the easiest differentiator in the build process as it can also be set to things like dragonwell, corretto, bisheng which are all fundamentally hotspot based, as well as openj9 which isn't.

> @sxa 'temurin-17-jdk' seems fine for me. But there is no temurin JVM variant. See #438

You're right that there is no difference in the JIT (other than anything arising from the compilers we use, and the compiler options) however Eclipse Adoptium is the vendor so it's right that the top level directory has that, but Temurin is our brand of openjdk so there's a good argument to have that in the directory path too IMHO to make the name visible, especially if it gives us consistency in the naming with the Linux installers.

(Note - I'm willing to be convinced either way, but I'm probably 80% towards going for temurin in the name and I'm interested to hear everyone's opinions on it :-) )

douph1:
What will happen to the running JVM when you delete/add/replace files/libs from same folder of existing process ?

steelhead31:
I'm assuming this is just a default path shown in the MSI installer, and can be changed, in the instance that I want to install multiple JDKs of the same major version?

douph1:
I think I remember that some time ago Oracle installer ask user to close running jvm before upgrade in place.
We can't do that (because we don't know how).

Why are older Oracle installer was doing that before ? To not have a running jvm with some new files and some old files, or because some files where locked and cannot be updated in place.
I think MSI can handle later upgrade of locked file on restart.. but then the goal is not achieved until restart. 

I think we must build some installer with major version (only) in path to allow to test upgrade while running jvm before merging into master.

Okeanos:
As the one who originally created https://github.com/adoptium/installer/issues/422 and looks very much forward to getting this solved, I am wondering what has to happen so this PR gets looked at & approved?

karianna:
Also see https://github.com/adoptium/installer/issues/983

karianna:
@adipiciu Would you be able to resolve the merge conflict / rebase?

sxa:
> Zulu, Liberica and Corretto are also using only the major version in the installdir path.

Noting that Red Hat's installer is the same as Temurin (It goes into `C:\Program Files\RedHat\java-21-openjdk-21.0.4.0.7-1` for the latest release), so there is precedent both ways

adipiciu:
@karianna Sorry, you can close this PR. I can't wait 1 year for these guys to take a decision. 
@Okeanos You can use msiexec /i "msifilename.msi" INSTALLDIR="C:\Program Files\Eclipse Adoptium\jre-8-hotspot" or whatever fixed name you want. Or even use Zulu, Corretto or Liberica builds which already have fixed paths.

Okeanos:
@adipiciu thanks for initially working on this first of all and thank you also for your suggestion. I know that I can override the default location when invoking the installer (and routinely do so). My goal was to make all of this smoother for automated processes, especially considering e.g. `winget` where it is not guaranteed that the Temurin manifest exposes this switch all the time.

Would it be okay for me to reuse the changes you proposed in your PR to create a new one? I would attribute these changes to you, of course.

@karianna, @douph1, @sxa could you give us a rough outline here what still has to be decided/what official outcomes are which need to be incorporated into a new PR?

karianna:
> @adipiciu thanks for initially working on this first of all and thank you also for your suggestion. I know that I can override the default location when invoking the installer (and routinely do so). My goal was to make all of this smoother for automated processes, especially considering e.g. `winget` where it is not guaranteed that the Temurin manifest exposes this switch all the time.
> 
> Would it be okay for me to reuse the changes you proposed in your PR to create a new one? I would attribute these changes to you, of course.
> 
> @karianna, @douph1, @sxa could you give us a rough outline here what still has to be decided/what official outcomes are which need to be incorporated into a new PR?

That would be welcome - I think given #983 let's start with what was proposed here: https://github.com/adoptium/installer/pull/789#issuecomment-1941924631

PR #790, opened 2024-01-07
Updated wix build documentation

adipiciu:
Updated wix3.11 and W10 sdk links
Added missing WIX variable
Removed calling the batch file from step 3, because it's done at step 4


karianna (approved):

gdams:
we're in the process of moving to wixV5 and https://github.com/adoptium/installer/pull/851 already provides this updated documentation

Issue #732, opened 2023-08-04
MSI-Installer: Text is cut off and not readable


larsen0815:
### Please provide a brief summary of the bug

The help text for at least one element that can be installed is cut off (notice the "..."). 

![Clipboard-1](https://github.com/adoptium/adoptium-support/assets/2345512/9173901e-2a4b-42cc-ba4c-1bc70954ab77)

### Please provide steps to reproduce where possible

- On Windows, start the installer for Temurin JDK 17
- Continue until the page where it is possible to select elements to be installed
- Select "JavaSoft (Oracle) registry keys"
- See text on the right

### Expected Results

Text should be short enough to be fully readable. Or better: Provide a link to a web page that has more information.

### Actual Results

Text is cut off (to be precise: something is left out from the middle) and in this form makes no sense at all.

### What Java Version are you using?

17

douph1:
As noted here 
https://github.com/adoptium/installer/blob/ce1b4858058ff90ca5866db66274d05f780170f7/wix/Lang/OpenJDK.Base.fr-fr.wxl.template#L19

Description must be limited to 255 chars (spec that I found somewhere one time) .

This one on en-us is  258 chars after substituting {vendor_branding}) by "Eclipse Temurin"
In fr-fr around 256 with {vendor_branding}) replaced by "Eclipse Temurin"

I think I remember that the full text is show when the cursor go in top of the text.

Feel free to propose a shorter description for each language to fit into 255 chars

adipiciu:
There is even a validation warning when building the msi:
warning SMOK1076 : ICE03: String overflow (greater than length permitted in column); Table: Feature, Column: Description, Key(s): FeatureOracleJavaSoft

![image](https://github.com/adoptium/installer/assets/13155063/02af4e7a-0135-4e06-82d0-0bfdfb8c2d49)

adipiciu:
I shortened the text under 255 length but it still doesn't show properly.
 ![image](https://github.com/adoptium/installer/assets/13155063/08c48b91-8961-46a0-b76b-a80042774b8c)

adipiciu:
I propose to shorten the string, so it can fit and also to get rid of the validation warning.
If the string cannot fit, it adds ellipsis (...) and you won't understand anything anyway.

@douph1 what do you say?
![image](https://github.com/adoptium/installer/assets/13155063/ce32c4ad-339e-416b-9a7e-673e35fb3daa)

karianna:
[...]

PR #791, opened 2024-01-09
Modify the feature description to fit in the installation window #732

adipiciu:
Modify the feature description to fit in the installation window #732

### DIFF
--- a/wix/Lang/OpenJDK.Base.en-us.wxl.template
+++ b/wix/Lang/OpenJDK.Base.en-us.wxl.template
@@ -17,7 +17,7 @@
 	<String Id="FeatureJarFileRunWithDescription">Associate .jar files to run with {vendor_branding}</String>
 	<String Id="DowngradeDenied">"A later version of [ProductName] is already installed. Setup will now exit."</String>
 	<String Id="FeatureOracleJavaSoftName">JavaSoft (Oracle) registry keys</String>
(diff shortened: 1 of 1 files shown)


wix/Lang/OpenJDK.Base.en-us.wxl.template
```diff
@@ -17,7 +17,7 @@
 	<String Id="FeatureJarFileRunWithDescription">Associate .jar files to run with {vendor_branding}</String>
 	<String Id="DowngradeDenied">"A later version of [ProductName] is already installed. Setup will now exit."</String>
 	<String Id="FeatureOracleJavaSoftName">JavaSoft (Oracle) registry keys</String>
-	<String Id="FeatureOracleJavaSoftDescription">Overwrites the reg keys HKLM\Software\JavaSoft (Oracle). After uninstallation of {vendor_branding}, Oracle Java running from PATH "C:\Program Files (x86)\Common Files\Oracle\Java\javapath" does not work. You need to reinstall it to recreate these registry keys</String>
+	<String Id="FeatureOracleJavaSoftDescription">Overwrites the reg keys HKLM\Software\JavaSoft (Oracle). After uninstallation of Temurin, Oracle Java needs to be reinstalled to re-create these registry keys.</String>
```
karianna, line 20: I think you need to keep {vendor_branding} for other folks who use these scripts (MSFT and IBM for example)

karianna (changes requested):

wix/Lang/OpenJDK.Base.en-us.wxl.template
adipiciu: If we keep {vendor_branding}, we need to make the string even smaller.

adipiciu: I have updated the text
image

karianna (approved):

douph1 (approved):

Issue #792, opened 2024-01-12
SLES repository does not incude 15.5


sxa:
Users on SLES 15 SP5 who try to follow the installation instructions at https://adoptium.net/installation/linux/#_opensusesles_instructions will find that they cannot download the Temurin packages because it does not include a directory suitable for the situation where VERSION_ID=15.5 - only 15.1 through 15.4 are in place at https://packages.adoptium.net/ui/native/rpm/opensuse

We should ensure that we publish a version for 15.5 so that the documentation works as expected on that distribution.

PR #793, opened 2024-01-12
Add OpenSuSE 15.5 as a supported distribution

sxa:
This is in draft while it's undergoing testing, but this adds support for OpenSuSE 15.5 - the current verison - into our packaging process to ensure that users using that distribution can follow the installation instructions without modification on the latest patch level. Note that this also adjusts the process to build the package on OpenSuSE 15.5 instead of 15.4 as the base image.

Potential alternative options:
- Add 15.6 which is available alpha for OpenSuSE 
- Remove 15.3 (out of support Dec 2022) to keep the number of distributions at 2, and potentially also rmove 15.4 since that is also out of support as of December 2023

Draft until I ensure it can be built successfully.

Fixes https://github.com/adoptium/installer/issues/792

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -464,6 +464,7 @@ def uploadRpmArtifacts(String DISTRO, String rpmArch) {
         'suse'   : [
             'rpm/opensuse/15.3',
             'rpm/opensuse/15.4',
+            'rpm/opensuse/15.5',
             'rpm/sles/12',
             'rpm/sles/15'
         ]
--- a/linux/README.md
+++ b/linux/README.md
@@ -229,6 +229,7 @@ SRPM also available.
 | oraclelinux/8 |         x86_64         |                                             |
 | opensuse/15.3 |         x86_64         |                                             |
 | opensuse/15.4 |         x86_64         |                                             |
+| opensuse/15.5 |         x86_64         |                                             |
 | rocky/8       |         x86_64         |                                             |
 | rpm/rhel/7    |         x86_64         |                                             |
 | rpm/rhel/8    |         x86_64         |                                             |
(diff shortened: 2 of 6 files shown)


sxa:
Failure when trying to test this as per https://adoptium.slack.com/archives/C9C272Z1T/p1705070159433229 even with the :
```
packaging.ZypperOperationsTest.packageSuccessfullyInstalled(String, String)[1] STANDARD_ERROR
    SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
    SLF4J: Defaulting to no-operation (NOP) logger implementation
    SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
but since the process seemed to create the `rpm` file successfully with this change and I've verified they install ok on 15.4 on aarch64 I'm going to mark it as ready for review.

karianna (approved):

karianna:
@sxa does this need merging for the release?

sxa:
> @sxa does this need merging for the release?

I'm always nervous about putting changes in this close so I@ve been wavering, but I think it makes sense in this case since I'd expect most people using that distribution to be on 15SP5. 

I believe @steelhead31 [has confirmed](https://adoptium.slack.com/archives/C9C272Z1T/p1705081075078739?thread_ts=1705070159.433229&cid=C9C272Z1T) that it seems to run through out processes ok ([Link](https://ci.adoptium.net/job/adoptium-packages-linux-pipeline_new/940/console)) and I've confirmed that the built package can still install on 15.3  so my gut feel is to try and get it in now I reckon.

smlambert (approved):

sxa:
Thank you @karianna!

PR #812, opened 2024-01-29
Reinstate libfreetype install on JDK21

sxa:
Fixes https://github.com/adoptium/temurin-build/issues/3624#issuecomment-1915272320 (hopefully!)

Note that I haven't adjusted `redhat/src/main/packaging/microsoft/21/msopenjdk-21.spec` which still has the `rm`

I have not tested this, so if a review can fdire off a test and see if the package contains `libfreetype.so` after this then that would be appreciated.

PR #824, opened 2024-02-22
Bump RPM spec & Debian versions to JDK21.0.2.0.0+13-2

### LINKED ISSUE #823: Republish JDK21.0.2.0.0+13 Deb & RPMs With Freetype Library Fix
The Debian & RPM installers for the most recent version of JDK21 require a republish/update with the fix made in #812 to include the freetype installer.

steelhead31:
Fixes #823 

Bump the RPM spec, and the debian version numbers to allow an update to include the freetype library fix, made as part of https://github.com/adoptium/installer/pull/812

### DIFF
--- a/linux/jdk/debian/src/main/packaging/temurin/21/debian/changelog
+++ b/linux/jdk/debian/src/main/packaging/temurin/21/debian/changelog
@@ -1,15 +1,21 @@
+temurin-21-jdk (21.0.2.0.0+13-2) STABLE; urgency=medium
+
+  * Eclipse Temurin 21.0.2.0.0+13-2 release.
+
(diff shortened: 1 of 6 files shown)


jiekang:
Seems fine to me. Would be nice to have someone with more Debian expertise to understand the implications of the change. I wonder if there is a debian tool similar to rpmdev-vercmp that you can just put in the two numbers and see what the result is.

E.g. with vercmp, you can check which version would be 'higher' between
`21.0.2.0.0+13-2` & `21.0.2.0.0+13`

As long as the `-2` version is considered higher (by Debian), we should be good.

jiekang:
Nice, thanks for checking. LGTM.

karianna (approved):

sebastianblesgen (approved):
LGTM

andrew-m-leonard (approved):

PR #833, opened 2024-03-04
User can now specify regexes used to find binaries

jmjaffe37:
Previously used regexes are now default parameter values. Also replaced some unwieldy cmd code with powershell script


wix/Build.OpenJDK_generic.cmd
```diff
@@ -147,6 +120,8 @@ FOR %%A IN (%ARCH%) DO (
     IF !PLATFORM! == x86-32 (
         SET PLATFORM=x86
     )
+    @REM Microsoft uses "x86" instead of "x86-32" for the platform folder
+    IF !VENDOR! == Microsoft SET FOLDER_PLATFORM=!PLATFORM!
```
karianna, line 124: Is this something we could set as a default property but be able to override inject that property directly instead of having to do the VENDOR check inside here?

wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
```diff
@@ -1,10 +1,51 @@
...
+    A regular expression that matches the JVM. Default is (?<jvm>hotspot|openj9|dragonwell).
+
+.PARAMETER jvm
+    The JVM to be used. If not provided, the script will attempt to extract the JVM from the filename.
+
+.PARAMETER platform_regex
+    A regular expression that matches the platform. Default is (?<platform>x86-32|x64|aarch64).
+
+.NOTES
+    File Name: CreateSourceFolder.AdoptOpenJDK.ps1
```
karianna, line 22: Would probably prefer to use `EclipseAdoptium` or `Adoptium` but I suspect `AdoptOpenJDK` has meaning, or would break if we changed it or....

karianna (approved):

wix/Build.OpenJDK_generic.cmd
jmjaffe37: I could add `x86` as an additional valid platform. Would we prefer that? @karianna @gdams

wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
karianna: That's what I suspected, ignore my request!

karianna (approved):

gdams (approved):

wix/Build.OpenJDK_generic.cmd
```diff
@@ -36,55 +36,28 @@ IF NOT DEFINED PRODUCT_HELP_LINK SET PRODUCT_HELP_LINK=https://github.com/adopti
...
-            )
-        )
-	)
+powershell -ExecutionPolicy Bypass -File "%~dp0\helpers\Validate-Input.ps1" ^
+    -toValidate '%ARCH%' ^
+    -validInputs "x64 x86-32 x86 arm64" ^
+    -delimiter " "
+
+IF %ERRORLEVEL% == 1 (
+    ECHO ARCH %ARCH% not supported : valid values are any combination of : x64, x86-32, arm64
```
douph1, line 45: "x86" is missing here.

(I don't know how x86 is different from x86-32 and why it as been added)

jmjaffe37: `x86` is the same as `x86-32`. Some vendors (like Microsoft) use `x86` instead of `x86-32` so this change helps vendors use this code without local modification

jmjaffe37: Note, I updated this for you in my latest PR:
https://github.com/adoptium/installer/pull/851

PR #838, opened 2024-03-12
JDK22 Installer Spec Files

adamfarley:
The creation of the installer spec files for the March 2024 JDK22 release.


adamfarley:
Basing this change set on the changes made for jdk21, plus the subsequent change sets that appear relevant.

https://github.com/adoptium/installer/commit/0ca368e31d17850808f7a1b717d140dd85d0a0c9

Also added a few changes to code that tests for jdk21, as it seems the supporting reasoning applies to subsequent JDK versions as well.

adamfarley:
Expected to fail many checks, as the hashes and build numbers are placeholders only, and don't match against anything.

karianna (approved):

andrew-m-leonard (approved):

PR #842, opened 2024-03-16
Update Main.wxs.template

adipiciu:
Removed the JavaSoftCurrentVersion component condition.

Temurin 8 JavaSoftCurrentVersion key is not removed at uninstall, because the condition is never satisfied. For Temurin 11, 17, etc, the condition doesn't make any sense because it's comparing with version "1.8" and it's always true, and Temurin 8 has totally different keys anyways.

### DIFF
--- a/wix/Main.wxs.template
+++ b/wix/Main.wxs.template
@@ -133,7 +133,6 @@
         <RegistryValue Root="HKMU" Key="SOFTWARE\Classes\{vendor}.jarfile\shell\open\command" Type="string" Value="&quot;[INSTALLDIR]bin\javaw.exe&quot; -jar &quot;%1&quot; %*" KeyPath="no" />
       </Component>
       <Component Id="SetOracleJavaSoftKeysCurrentVersion" Guid="*">
-          <Condition><![CDATA[NOT JAVASOFT_CURRENTVERSION OR JAVASOFT_CURRENTVERSION <> "1.8" AND JAVASOFT_CURRENTVERSION < $(var.ProductMajorVersion)]]></Condition>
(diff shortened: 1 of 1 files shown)


douph1:
The condition was added to prevent overriding CurrentVersion with last version over higher version.

As mentioned in comment here :
https://github.com/adoptium/installer/issues/329#issuecomment-900660819

>I have fixed the wrong comparaison order "<" in 26d01e8e818d0b956e86e9746268015d11904319 **but not fixed the jdk8 specific case**

I think we must handle the 1.8 case not remove the logic completly.
The actual part of the condition "JAVASOFT_CURRENTVERSION <> "1.8" " is probably making Wix to completly ignore
then <RegistryValue ... /> when JDK 1.8 is installed.


https://github.com/adoptium/installer/issues/329

>For Temurin 11, 17, etc, the condition doesn't make any sense because it's comparing with version "1.8" and it's always true

The condition skip comparing existing installed ( (1.8)left <>)  with installer version.

Maybe we must add another test for ignoring (<> right (1.8) )  : $(var.ProductMajorVersion) <> 1.8
or just handle "1.8" CurrentVersion after.


With other than "1.8" version, numeric compare is done to not lower CurrentVersion with older version when you install old java after last version.

Goal : 
Install Java 21 > CurrentVersion = 21
Then
Install Java 17 > CurrentVersion = 21

adipiciu:
I agree for Temurin 11, 17, etc. Then the condition should be changed to:

NOT JAVASOFT_CURRENTVERSION OR JAVASOFT_CURRENTVERSION = $(var.OracleVersionKey)

The condition is evaluated both at install and uninstall.

At install -> it will write the registry key if it's null, or if the value is the same with the installer version key.
At uninstall -> it will remove the key if it's null, or if the value is the same with the installer version key.

So, the logic is that it will remove the key only if the value is identic to the temurin installer version (8, 11, 17, etc.)


Btw, the var.OracleVersionKey value is 8, not 1.8. So the condition was wrong from the start.

PR #844, opened 2024-03-21
Update JDK22 Installer files for JDK22+36 Release

steelhead31:
These files have s390x removed due to some blocking tests, this PR will allow publishing of all other linux installers. A 2nd PR to publish the s390x installers ( and an additional one to come will also be made, to include riscv support )

### DIFF
--- a/linux/jdk/debian/src/main/packaging/temurin/22/debian/rules
+++ b/linux/jdk/debian/src/main/packaging/temurin/22/debian/rules
@@ -8,8 +8,8 @@ amd64_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/down
 amd64_checksum = bc3d99e816d0c373f424cd7aa2b6d3e8081a7189fe55c1561616922200ec8e47
 arm64_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22%2B36/OpenJDK22U-jdk_aarch64_linux_hotspot_22_36.tar.gz
 arm64_checksum = 4b52670caea44848cee893e35c804380817b6eff166cf64ee70ca2bfaac3d1c7
-TO_DO_ppc64el_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22.0.0%2B0/OpenJDK22U-jdk_ppc64le_linux_hotspot_22.0.0_0.tar.gz
-TO_DO_ppc64el_checksum = d08de863499d8851811c893e8915828f2cd8eb67ed9e29432a6b4e222d80a12f
+ppc64el_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22%2B36/OpenJDK22U-jdk_ppc64le_linux_hotspot_22_36.tar.gz
+ppc64el_checksum = 8c062e934d95c639f97b4e51b968eed694a6653248727c3db8bc5e0e55cfd7f4
 
 
 d = debian/$(pkg_name)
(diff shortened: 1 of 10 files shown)


linux/jdk/debian/src/main/packaging/temurin/22/debian/rules
```diff
@@ -3,14 +3,14 @@
...
 jvm_tools = jar jarsigner java javac javadoc javap jcmd jconsole jdb jdeprscan jdeps jfr jhsdb jimage jinfo jlink jmap jmod jpackage jps jrunscript jshell jstack jstat jstatd jwebserver keytool rmiregistry serialver jexec jspawnhelper
-amd64_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22.0.0%2B0/OpenJDK22U-jdk_x64_linux_hotspot_22.0.0_0.tar.gz
-amd64_checksum = 454bebb2c9fe48d981341461ffb6bf1017c7b7c6e15c6b0c29b959194ba3aaa5
-arm64_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22.0.0%2B0/OpenJDK22U-jdk_aarch64_linux_hotspot_22.0.0_0.tar.gz
-arm64_checksum = 3ce6a2b357e2ef45fd6b53d6587aa05bfec7771e7fb982f2c964f6b771b7526a
-ppc64el_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22.0.0%2B0/OpenJDK22U-jdk_ppc64le_linux_hotspot_22.0.0_0.tar.gz
-ppc64el_checksum = d08de863499d8851811c893e8915828f2cd8eb67ed9e29432a6b4e222d80a12f
-s390x_tarball_url = https://github.com/adoptium/temurin22-binaries/releases/download/jdk-22.0.0%2B0/OpenJDK22U-jdk_s390x_linux_hotspot_22.0.0_0.tar.gz
-s390x_checksum = 0d5676c50821e0d0b951bf3ffd717e7a13be2a89d8848a5c13b4aedc6f982c78
+
```
sxa, line 6: Is this new blank line intentional. That's really confused git's diff view ;-)

linux/jdk/redhat/src/main/packaging/temurin/22/temurin-22-jdk.spec
```diff
@@ -3,10 +3,10 @@
 # https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_upstream_uses_invalid_characters_in_the_version
 # also not very intuitive:
 #  $ rpmdev-vercmp 22.0.0.0.0___22.0.0.0.0+1
-#  22.0.0.0.0___1 == 22.0.0.0.0+35
-%global spec_version 22.0.0.0.0.0
+#  22.0.0.0.0___1 == 22.0.0.0.0+36
+%global spec_version 22.0.0.0.0.36
 %global spec_release 1
-%global priority 1161
+%global priority 2200
```
sxa, line 9: Where does this number come from? They don't seem to be consistent across versions.

sxa (approved):
I'll approve this on the basis that if it doesn't work you know how to clear it up ;-)

linux/jdk/debian/src/main/packaging/temurin/22/debian/rules
steelhead31: I think it was a legacy of removing s390x, I've tidied this up, which should help git.

linux/jdk/redhat/src/main/packaging/temurin/22/temurin-22-jdk.spec
steelhead31: To the best of my knowledge, it has to be the same for all versions of a package. I don't think we actually use it anywhere, but maybe @jiekang can offer some insight ?

jiekang: It's used for the priority value in alternatives, a Linux tool for selecting the JDK to use for a system. For example here: https://github.com/adoptium/installer/blob/master/linux/jdk/redhat/src/main/packaging/temurin/21/temurin-21-jdk.spec#L166

These comments and usage in the Fedora spec share some insight:
https://src.fedoraproject.org/rpms/java-21-openjdk/blob/rawhide/f/java-21-openjdk.spec#_339

It looks like they should mimic the jdk version and be updated when releasing new versions.

tellison (approved):

gdams (approved):

PR #846, opened 2024-03-25
Syntax Fix For Version Compare.

steelhead31:
Fix missing single quotes for Alpine version compare.

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -253,10 +253,10 @@ def jenkinsStepAlpine() {
         apkArchAllList = ['x86_64', 'aarch64']
     }
     // remove aarch64 for JDK20 and below
-    if (${VERSION} -lt 21) {
+    if (${VERSION} < '21') {
         apkArchAllList.remove('aarch64')
     }
-
+    
     apkArchAllList.each { ApkARCH ->
         // special handle: no label x86_64 only x64 for alpine agent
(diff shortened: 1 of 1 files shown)


linux/Jenkinsfile
```diff
@@ -253,10 +253,11 @@ def jenkinsStepAlpine() {
         apkArchAllList = ['x86_64', 'aarch64']
     }
     // remove aarch64 for JDK20 and below
-    if (${VERSION} -lt 21) {
+    def versionInt = VERSION as int // Parse Version As Integer
```
zdtsw, line 256: isnt it better to do
`def versionInt = VERSION.toInteger()`
instead?

gdams (approved):

andrew-m-leonard (approved):

linux/Jenkinsfile
andrew-m-leonard: I think this is subjective, from a quick google "as Integer" seems the favorite, unless there's a specific reason?

steelhead31: I tried this, and it caused gradle to break, so I've reverted it back to my original format.

steelhead31:
successful run here : https://ci.adoptium.net/job/adoptium-packages-linux-pipeline_new/1272/

sxa (approved):

PR #107, opened 2019-05-07
enable capability to install per user + doc

douph1:
To close https://github.com/AdoptOpenJDK/openjdk-installer/issues/103

PR #851, opened 2024-04-10
Updated msi windows installers to use wix v5

jmjaffe37:
In moving to wix v5, some of the code could be simplified a bit (mainly: `light.exe` and `candle.exe` commands could be combined into one). For end users to use this wix installer creation tool, they will just need to run:
`dotnet tool install --global wix --version 5.0.0`

To install wix `5.0.0` (or they can install another version of wix v4 or wix v5 if they so choose).


jmjaffe37:
Link to successful run on my fork (although I believe that this PR will also do the same checks): https://github.com/jmjaffe37/openjdk-installer/actions/runs/8624015173

jmjaffe37:
Converted to draft due to ongoing release. Note: currently passing all PR checks

jmjaffe37:
Note: I had to change the xml propery ID `ALLUSERS` in `Main.wxs.template` to `ALLUSERSSETUP` to avoid a duplicate symbol error that I was getting locally. Here is a link to a run I did on github actions to recreate this error that I previously had locally:
https://github.com/jmjaffe37/openjdk-installer/actions/runs/8635538967/job/23673480230#step:7:61

jmjaffe37:
Just did a test run using wix `5.0.0` and it seems to have built properly:
https://github.com/jmjaffe37/openjdk-installer/actions/runs/8635971450

I also did an internal test run to create Microsoft OpenJDK MSIs and the resulting MSIs passed the test outlined above (install, correct java version, uninstall as expected)

jmjaffe37:
Is anyone able to help me test if the installer translations into other languages/cultures is still behaving as expected? I am unsure on how to do that at the moment 😅

karianna:
I think some useful testing will be to upgrade from older installations of Eclipse Temurin and MSFT builds of OpenJDK respectively.

karianna:
> Is anyone able to help me test if the installer translations into other languages/cultures is still behaving as expected? I am unsure on how to do that at the moment 😅

@douph1 - Are you able to check for French?

For the rest I recommend adding a note to the installer channel in Slack and giving folks some instructions on how to self serve this PR and test translations.

jmjaffe37:
> > Is anyone able to help me test if the installer translations into other languages/cultures is still behaving as expected? I am unsure on how to do that at the moment 😅
> 
> @douph1 - Are you able to check for French?
> 
> For the rest I recommend adding a note to the installer channel in Slack and giving folks some instructions on how to self serve this PR and test translations.

@karianna I actually figured out how to do it after posting that comment 😅
I updated it to instead explain how I tested it (I made that update before I saw your reply)

wix/Build.OpenJDK_generic.cmd
```diff
@@ -194,10 +201,9 @@ FOR %%A IN (%ARCH%) DO (
 
 
     REM Build with extra Source Code feature (needs work)
```
jmjaffe37, line 203: I updated the code in the comments below to be in the style of wix v4 and wix v5, but it seems that these comments have been untouched for a while now. Would it be better to delete them?

wix/BuildSetupTranslationTransform.cmd
```diff
@@ -13,12 +13,12 @@ SET LANGIDS=%LANGIDS%,%LANGID%
 
 ECHO Building setup translation for culture "%1" with LangID "%2"...
 REM Build with extra Source Code feature (needs work)
```
jmjaffe37, line 15: Same comment here:

I updated the code in the comments below to be in the style of wix v4 and wix v5, but it seems that these comments have been untouched for a while now. Would it be better to delete them?

jmjaffe37:
Another note: wix v5 was officially release (taken out of beta) on april 5th according to this blog post on their website:
https://www.firegiant.com/blog/2024/4/5/wix-v5.0.0-has-been-released/

Since I did my testing on both wix v4.0.5 and wix v5.0.0 (the first non-beta version), would we like me to change the default wix version to 5.0.0?

karianna:
> Another note: wix v5 was officially release (taken out of beta) on april 5th according to this blog post on their website: https://www.firegiant.com/blog/2024/4/5/wix-v5.0.0-has-been-released/
> 
> Since I did my testing on both wix v4.0.5 and wix v5.0.0 (the first non-beta version), would we like me to change the default wix version to 5.0.0?

Yes please

wix/BuildSetupTranslationTransform.cmd
karianna: I would remove if they're not providing clarity

jmjaffe37: These comments contain code that, in theory, someone could uncomment and run to "Build with extra source code." That said, they had not been touched in 6 years and seemed to be out of date anyways. I did my best here to update them to what is currently being used, but it also feels like we can just retire the dead code. If nobody protests then I will remove these comments in the code in a few days

wix/Main.wxs.template
```diff
@@ -1,238 +1,232 @@
...
-  <?include $(sys.CURRENTDIR)\Workdir\$(var.OutputBaseFilename)-$(var.ProductSku).Variables.wxi ?>
-
-  <Product Id="$(var.ProductId)" Name="$(var.ProductNameWithVersion)" Language="$(var.ProductLanguage)" Version="$(var.ProductVersion)" Manufacturer="$(var.ProductManufacturer)" UpgradeCode="$(var.ProductUpgradeCode)">
-    <Package Description="$(var.PackageDescription)" Manufacturer="$(var.ProductManufacturer)" InstallerVersion="200" Compressed="yes" />
-    <MajorUpgrade DowngradeErrorMessage="!(loc.DowngradeDenied)"/>
-    <Media Id="1" Cabinet="Data1.cab" EmbedCab="yes" CompressionLevel="high" />
-    <Property Id="ARPHELPLINK" Value="!(loc.ProductHelpLink)" />
-    <Property Id="ARPURLINFOABOUT" Value="!(loc.ProductUrlInfoAbout)" />
-    <Property Id="ARPURLUPDATEINFO" Value="!(loc.ProductUrlUpdateInfo)" />
-    <Property Id="ALLUSERS" Value="2" /> <!-- Allow per User or per Machine setup -->
```
jmjaffe37, line 12: @gdams @douph1 I had to change the name of the property `ALLUSERS` to avoid a duplicate symbol error. I also changed all references to this property, but I now see that this property is an  actual reserved property name that MSI uses. I am still looking for a solution that allows me to set the `ALLUSER` property (the SetProperty tag didn't seem to work), but I wanted to ask if there were any other ideas

wix/helpers/Update-id.ps1
```diff
@@ -0,0 +1,72 @@
...
+            $_ -replace $('Directory="' + [regex]::Escape($directoryId) + '"'), $('Directory="' + [regex]::Escape($updatedDirectoryId) + '"')
+        }
+    }
+
+    # Write the updated content back to the file
+    $updatedContent | Set-Content -Path $FilePath
+}
+
+
+
```
gdams, line 72: extra newlines here

wix/BuildSetupTranslationTransform.cmd
gdams: yeah I'd remove them

wix/Main.wxs.template
gdams: @jmjaffe37 have you asked upstream? Might be worth discussing this with the wix maintainers

.github/workflows/wix.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           TAG: jdk8u362-b09
           SUB_TAG: 8u362b09
           JVM: hotspot
+          WIX_VERSION: 5.0.0
```
gdams, line 27: guessing this isn't strictly needed if you're already setting the default to be 5.0.0? Maybe remove these and the `-wix_version ${{ env.WIX_VERSION }}` flag

wix/Main.wxs.template
jmjaffe37: @gdams, Is there a slack channel I could ask in? Or should I open an issue in the wix github repo: https://github.com/wixtoolset/wix

wix/Build.OpenJDK_generic.cmd
```diff
@@ -80,6 +82,10 @@ SET WIN_SDK_FULL_VERSION=10.0.17763.0
 SET WORKDIR=Workdir\
 mkdir %WORKDIR%
 
+@REM Add necessary wix extensions here
```
gdams, line 85: I think my preference here would be to install these at an infra level rather than everytime. Perhaps add this to the GitHub workflow and then I'll make sure that the extensions exist on our CI machines along with the updated wix V5?

.github/workflows/wix.yml
```diff
@@ -57,6 +60,7 @@ jobs:
       - name: Install dependencies
         run: |
           choco install --no-progress wget
```
gdams, line 62: somewhat unrelated but I don't think this is required anymore. As long as we use `pwsh.exe`

.github/workflows/wix.yml
jmjaffe37: I see you point. I originally though that this could be a nice way to let developers know that this is an option (just by looking at the yaml), and that both scripts need the wix_version input. If we think it is a better option, I can just leave this info in the README.md instead

gdams: yeah lets keep the workflow simple and document any additional configuration options in the README.md

wix/Build.OpenJDK_generic.cmd
jmjaffe37: I can try this, although I have 2 thoughts:
1) Doing this at the infra level means that all vendors would need to read the README and add this themselves or the code will break
1) This command adds the extensions to a local, project specific subdirectory. I would believe this would mean that most build scenarios would have to install this every time anyways

.github/workflows/wix.yml
jmjaffe37: Ok, I can try a run where I remove the choco install command

jmjaffe37:
> @jmjaffe37 great work getting this sorted! I'll take a more thorough review this week. One note though is that it kinda sucks that we've lost the `<?xml version="1.0" encoding="utf-8"?>` definition at the top of the files as it means that GitHub no longer highlights the XML files for us. Does v4 not support this at the top?

I can try a run with this added back. I believe the inclusion of this line is optional for wix v4 and v5

wix/BuildSetupTranslationTransform.cmd
jmjaffe37: Done

gdams:
> I can try a run with this added back. I believe the inclusion of this line is optional for wix v4 and v5

There's certainly no harm to add it back if it's optional :)

wix/Build.OpenJDK_generic.cmd
gdams: ahh I hadn't realised that was the case. In which case maybe it's better staying where it is

.github/workflows/wix.yml
jmjaffe37: Just ran a test run, looks like the `choco install` is still necessary:
https://github.com/adoptium/installer/actions/runs/8788854487/job/24117265968?pr=851#step:6:21

.github/workflows/wix.yml
jmjaffe37: Done

wix/helpers/Update-id.ps1
jmjaffe37: Fixed

jmjaffe37:
> @jmjaffe37 great work getting this sorted! I'll take a more thorough review this week. One note though is that it kinda sucks that we've lost the `<?xml version="1.0" encoding="utf-8"?>` definition at the top of the files as it means that GitHub no longer highlights the XML files for us. Does v4 not support this at the top?

xml tags have been added back :)

wix/Main.wxs.template
jmjaffe37: I asked here this morning, will update when we find a solution:
https://github.com/orgs/wixtoolset/discussions/8151

jmjaffe37: Ended up finding out that it is equivalent to set the Package `Scope` to `perUserOrMachine`, fix has been implemented

wix/README.md
```diff
@@ -7,6 +7,8 @@
 
 ## How to upgrade to a new OpenJDK version
 
+0. Ensure the necessary dependencies are installed (`wget` and `wix` version 5.0.0 or higher). If you do not have wix installed, you can followe the instructions [here](https://wixtoolset.org/docs/intro/#nettool) to install it.
```
gdams, line 10: I'm now a huge fan of starting a `0`. If you replace them all with `1` markdown will handle the numbering automatically.

Also wget isn't actually a dependency IMO, we just need it in GitHub action to easily pull down some prebuilt binaries.

karianna: Was that meant to be _not_ a huge fan?

gdams: > Was that meant to be _not_ a huge fan?

yeah I edited it just before you added this comment :)

gdams:
My initial testing shows that the default install dir is wrong. See the new installer on the left vs the old installer on the right

Screenshot 2024-04-24 at 10 00 33

wix/Main.wxs.template
```diff
@@ -1,238 +1,233 @@
...
+      <InstallExecuteSequence>
+        <Custom Action="PreventDowngrading" After="FindRelatedProducts">NEWPRODUCTFOUND</Custom>
+        <RemoveExistingProducts After="InstallFinalize" />
+      </InstallExecuteSequence>
+      <InstallUISequence>
+        <Custom Action="PreventDowngrading" After="FindRelatedProducts">NEWPRODUCTFOUND</Custom>
+      </InstallUISequence>
+    -->
+
+    <!-- Define the directory structure -->
```
gdams, line 39: Guessing you need to move this comment to line up with it's new position at https://github.com/adoptium/installer/pull/851/files#diff-1a76603c3a0aced9dc527a38408ba3816cf0d48cb10487bdbe10dbddad0623a3R211

jmjaffe37:
> ```
> InstallPrivileges
> ```

I can look into this some more, but it looks like the `InstallPriveleges` attribute is no longer part of the package element in wix v4 or v5. See this page for a list of available attributes:
https://wixtoolset.org/docs/schema/wxs/package/#attributes

jmjaffe37:
> My initial testing shows that the default install dir is wrong. See the new installer on the left vs the old installer on the right
> 
> Screenshot 2024-04-24 at 10 00 33
> Doing some research I think this may be linked to `InstallPrivileges="elevated"` which can be defined at the `<Package />` definition although we don't want to break perUser installs so we may need some logic here
> 
> E.g:
> 
> ```
> <Package InstallerVersion="200" Compressed="yes" InstallScope="perMachine" InstallPrivileges="elevated" />
> ```

I can look into this some more, but it looks like the InstallPriveleges attribute is no longer part of the package element in wix v4 or v5. See this page for a list of available attributes:
https://wixtoolset.org/docs/schema/wxs/package/#attributes

wix/Main.wxs.template
jmjaffe37: There doesn't seem to be a line number attahed to this link. If I had to guess, maybe you are asking for the comment on line 39 to be moved to line 19?

jmjaffe37:
Note: to address a request made here: https://github.com/adoptium/installer/issues/888, I implemented a feature to allow vendors to implement custom licenses for hotspot JVMs. It also allows the vendors to choose whether or not to have the end user read and agree to this license (for example, this is optional for the GPLv2 license)

Successful test run with visible GPLv2 license for `Eclipse Adoptium` vendor: https://github.com/adoptium/installer/actions/runs/8902373183?pr=851

See the PR check for a successful test run for `Eclipse Adoptium` vendor MSIs without the license showing (default behavior for all vendors with the GPLv2 license unless specified)

gdams:
raised https://github.com/adoptium/infrastructure/pull/3545 to install wix v5.0.0 on the CI machines

jmjaffe37:
Ok, I believe that this PR is ready to be merged now!

CC: @gdams, @karianna, @douph1

wix/Main.wxs.template
```diff
@@ -178,51 +178,51 @@
...
-        <RegistryValue Root="HKLM"
-          Key="SOFTWARE\JavaSoft\$(var.OracleJavasoftBaseKey)\$(var.OracleVersionKey)"
-          Name="RuntimeLib" Type="string" Value="[INSTALLDIR]$(var.DllPath)" KeyPath="no" />
-        <!-- Oracle dont set RuntimeLib for JDK .. because JRE is "private" JRE so no ?else? -->
+          <RegistryValue Root="HKLM"
+            Key="SOFTWARE\JavaSoft\$(var.OracleJavasoftBaseKey)\$(var.OracleVersionKey)"
+            Name="RuntimeLib" Type="string" Value="[INSTALLDIR]$(var.DllPath)" KeyPath="no" />
+          <!-- Oracle doesn't set RuntimeLib for JDK .. because JRE is "private" JRE so no ?else? -->
         <?endif?>
         <RegistryValue Root="HKLM"
```
karianna, line 186: Whitespace for this doesn't match your change above?

.github/workflows/wix.yml
```diff
@@ -12,49 +12,62 @@ permissions:
...
-          PRODUCT_MAINTENANCE_VERSION: 6
-          PRODUCT_PATCH_VERSION: 0
-          PRODUCT_BUILD_NUMBER: 10
-          MSI_PRODUCT_VERSION: 17.0.6.10
-          ARCH: x64
-          TAG: jdk-17.0.6+10
-          SUB_TAG: 17.0.6_10
-          JVM: hotspot
+          - jdk: 8
+            ICEDTEA_WEB_VERSION: "icedtea-web-1.8.6"
```
karianna, line 22:
```suggestion
            ICEDTEA_WEB_VERSION: "icedtea-web-1.8.8"
```

wix/InstallerChangelog.md
```diff
@@ -8,6 +8,11 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
+## 2024-05-02
+### Changed
+- Update to wix v5
+- Fixed windows Arm installation path
+
```
karianna, line 15:
```suggestion
- Update IcedTea-Web to 1.8.8

```

wix/README.md
```diff
@@ -1,27 +1,31 @@
 # Requirements for build environment
 
-1. [Windows Installer XML (WiX) toolset, 3.11 or later](http://wixtoolset.org/releases/)
-2. Install ["Windows SDK for Desktop C++ amd64 Apps" feature from Windows SDK 10](https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk) for building multi-lingual setups.
-3. Digital signature service if the MSI should be signed (optional). If you plan to sign the MSI, you need to install the Windows SDK 10 feature "Windows SDK Signing Tools for Desktops Apps".
-4. For reviewing the MSI setup or creating custom MST transforms you can install feature "MSI Tools" from Windows SDK 10 (optional).
+1. [Windows Installer XML (WiX) toolset, 4.0.0 or later (5.0.X recommended)](https://wixtoolset.org/docs/intro/#nettool)
```
karianna, line 3: Shall we just make 5.0.0 the min instead of 4.0.0?

karianna (changes requested):
Some nitpicks

wix/InstallerChangelog.md
gdams: I'd prefer to not call this out in the changelog. It's unrelated to the changes (only affect GitHub CI)

gdams (approved):

wix/README.md
jmjaffe37: Works for me!

wix/Main.wxs.template
jmjaffe37: @karianna I am unsure which section you are referring to about the whitespace mismatch. I will say, some of the whitespace tweaks were done by @gdams, but I wanted to change back the inner blocks of if-statements to be indented so the reader is more clear on which lines fall within the if-statement block

gdams: @jmjaffe37 feel free to tweak them to however you prefer

karianna (approved):

karianna:
Probably worth communicating this major change to the whole community and landing all of the related PRs in sync.  Then we should ask for volunteers to test various scenarios (fresh install, upgrade from previously installed version with old WIX installer, downgrade etc).

steelhead31 (approved):
Tested GHA built MSI installers for JDK8, JDK11, JDK17 & JDK21 on Windows x64, all behave as expected, other than an expected oddity with upgrade behaviour, which will work normally when this PR is merged. Approving on that basis.

douph1:
Hi @jmjaffe37  great work.
I have seen you have made best effort to maintain dual capability of installer to work perUser or perMachine.

I have added perUser installer capability in https://github.com/adoptium/installer/pull/107 

Some time later, Adoptium team have decided that perUser is not the way to go and won't be supported anymore:
See
https://adoptium.net/installation/windows/ 
>The installer is designed for use on a per-machine basis, **not per-user basis**, so you can have only one installation of the MSI on a machine for all users.

If ALLUSERS was here it was only because cleanup was not done by the dev who disable perUser capability..

jmjaffe37:
> Hi @jmjaffe37 great work. I have seen you have made best effort to maintain dual capability of installer to work perUser or perMachine.
> 
> I have added perUser installer capability in #107
> 
> Some time later, Adoptium team have decided that perUser is not the way to go and won't be supported anymore: See https://adoptium.net/installation/windows/
> 
> > The installer is designed for use on a per-machine basis, **not per-user basis**, so you can have only one installation of the MSI on a machine for all users.
> 
> If ALLUSERS was here it was only because cleanup was not done by the dev who disable perUser capability..

Thanks for letting me know @douph1! This raises a few questions for me:

1) Does this mean that additional work would be necessary to support a perUser installation? From my understanding, my update takes care of the necessary prerequisites to support this.

2) Do we prefer to turn off the perUser option?

3) I noticed that the linked page states that only windows x64 is supported, but Microsoft has been using this code to generate installers for x86 and aarm64 without issue. In fact, recent updates have specifically debugged aarm64 installer creation. Is there somewhere I can prupose new wording?

Thanks!

CC: @d3r3kk

karianna:
@jmjaffe37 For that last point (docs) - you can send a PR into the [adoptium.net ](https://github.com/adoptium/adoptium.net) repo.

douph1:
1. I think TSC don't want to support "perUser", I don't understand why, as it worked by the past. If changing we probably need a little more testing for install/upgrade/uninstall and maybe cohabitation with x64. I have no time for this.

2. 2. I have no opinion, nor interest, but thanks for asking :)

3.
Adoptium seems to not build x86 MSI installer after JDK 19. JDK 20/21/22 installer for x86 is missing. I don't know why but it is maybe the meaning of "Note: Windows installer packages are supported only on Windows x64 systems."
If someone has clue on this we can fix the docs of the website.

Adoptium don't build/test/publish arm64 (AArch64) MSI installer. So this is not documented on the Website, but you can add some wording on wix/README.md

PR #873, opened 2024-04-25
 fix cacert upload action

gdams:


sxa (approved):
Approving because it looks fairly sane and I trust the Scott knows enough about the commands being run here (which I don't) to have sanitised them ;-)

PR #885, opened 2024-04-27
linux: run tests on all files

gdams:
We've been missing certain issues with dependency upgrades that aren't being tested correctly with the automatic matrix generator. This PR switches it to just run all build/tests which we can comfortably do now we have more GH actions minutes.

Also applies a fix to resolve the breaking changes in v4 upload/download artifacts


tellison (approved):

steelhead31 (approved):
LGTM

PR #906, opened 2024-05-15
Update Spec Files For JDK17 RiscV Release

### LINKED ISSUE #905: GPG Signing Fails For RHEL Installer Packages
All, following the bump to Fedora 40 in the dockerfiles for the RHEL JDK & JRE RPM package builds, the process now fails with the following error when GPG signing is enabled ( the error doesnt present when signing is not enabled ):`gpg: enabled compatibility flags:`  
`gpg: using pgp trust model`  
`gpg: no running keyboxd - starting '/usr/libexec/keyboxd'`  
`gpg: waiting for the keyboxd to come up ... (5s)`  
`gpg: connection to the keyboxd established`  
`gpg: Note: database_open 134217901 waiting for lock (held by 9) ...`Having done some testing and investigation around this, the process works fine until Fedora 39, ( Fedora 38 doesn't exhibit this problem ) Known Issue in Fedora 39:  
[https://discussion.fedoraproject.org/t/gpg-hang-on-fedora-silverblue-39/103262](https://discussion.fedoraproject.org/t/gpg-hang-on-fedora-silverblue-39/103262) , which still appears to be present in Fedora 40.

Two potential fixes.. to keep using Fedora 40

1) Comment Out: use-keyboxd in .gnupg/common.conf.  
2) Remove any .lock files prior to running the sign  
  
Fix 1, doesnt work inside the docker container, Fix 2 does, and allows the rpms to be signed correctly  
  
Alternatively reverting the docker containers to use a Fedora 38 base image would also work.

steelhead31:
Also Fixes #905 

Update the specfiles and jenkinsfile for the release of JDK17 for riscv64.

Also reverts the docker image update for the rpm package build to Fedora 38 to avoid the issue with gpg signing.

### DIFF
--- a/linux/Jenkinsfile
+++ b/linux/Jenkinsfile
@@ -9,7 +9,7 @@ aarch64     aarch64     arm64       arm64
 ppc64le     ppc64le     ppc64le
 source      src         -           only for SRPM and no need specify as option or target
 s390x       s390x       s390x       only for jdk8+
-riscv64     riscv64     riscv64     only for JDK21+
+riscv64     riscv64     riscv64     only for JDK17+
 */
 
 env.NODE_LABEL = 'dockerBuild&&linux&&x64' // Default node and also used for build RedHat + Suse + Debian x64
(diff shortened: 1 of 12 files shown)


tellison (approved):

andrew-m-leonard (approved):
lgtm

PR #958, opened 2024-07-22
Updated linux configs for july 2024 psu

jmjaffe37:


gdams (approved):

karianna (approved):

karianna:
JUnit failed because of a previous run, there's some sort of order of operations at fault here as you can't re-trigger that check.

Issue #962, opened 2024-07-24
JRE RPM package is not signed


bebanowski:
Same situation as https://github.com/adoptium/installer/issues/611, only for a newer release 

Package temurin-17-jre-17.0.12.0.0.7-1.x86_64.rpm is not signed

Signature is missing here:
```
$ rpm -qpi temurin-17-jre-17.0.12.0.0.7-1.x86_64.rpm
Name        : temurin-17-jre
Version     : 17.0.12.0.0.7
Release     : 1
Architecture: x86_64
Install Date: (not installed)
Group       : java
Size        : 140814170
License     : GPLv2 with exceptions
Signature   : (none)
Source RPM  : temurin-17-jre-17.0.12.0.0.7-1.src.rpm
Build Date  : Wed 24 Jul 2024 10:57:12 BST
Build Host  : c795d3e09c70
Relocations : /usr/lib/jvm/temurin-17-jre
Packager    : Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>
Vendor      : Eclipse Adoptium
URL         : https://projects.eclipse.org/projects/adoptium
Summary     : Eclipse Temurin 17 JRE
Description :
Eclipse Temurin JRE is an OpenJDK-based runtime environment to execute
applications and components using the programming language Java.
```

Can this be signed so we don't have to bypass the GPG check?

karianna:
@jerboaa Are you able to confirm?

steelhead31:
@bebanowski & @waeller 

Would you mind re-downloading the file and trying again.. there was a slight issue when the files were being uploaded last evening..

This has subsequently been rectified, but I believe you may have caught the invalid version of the package..

```
cfryer@scfryer-thinkpadp1gen2:~/Downloads$ rpm -qpi temurin-17-jre-17.0.12.0.0.7-1.x86_64.rpm
Name        : temurin-17-jre
Version     : 17.0.12.0.0.7
Release     : 1
Architecture: x86_64
Install Date: (not installed)
Group       : java
Size        : 140809828
License     : GPLv2 with exceptions
Signature   : RSA/SHA256, Wed 24 Jul 2024 01:50:54 PM BST, Key ID 843c48a565f8f04b
Source RPM  : temurin-17-jre-17.0.12.0.0.7-1.src.rpm
Build Date  : Wed 24 Jul 2024 01:45:56 PM BST
Build Host  : c62056947c1d
Relocations : /usr/lib64/jvm/temurin-17-jre 
[...]

PR #964, opened 2024-07-25
JULY CPU - JDK21 & JDK22 & JDK17 Updates For Shipping Linux Packages

steelhead31:
Updates required for JDK21 & JDK22 Alpine, Debian, RH & SUSE packages.

Also includes a bump for JDK17 / JRE17 to bump the spec version for some missing signatures.


smlambert (approved):

sxa (commented):
Tricky to review every part of this in detail but I expect it's ok and we can correct if something goes wrong.
One comment though - are [the changes to build.sh](https://github.com/adoptium/installer/pull/964/files#diff-e6b9dd1bc6d6f4e60e402932a3306783e062cc8683f364a7fc94bf238cacbab9) supposed to be in this PR? If so probably worth adding a comment into this PR mentioning it since it looks like a change in functionality.

steelhead31:
> Tricky to review every part of this in detail but I expect it's ok and we can correct if something goes wrong. One comment though - are [the changes to build.sh](https://github.com/adoptium/installer/pull/964/files#diff-e6b9dd1bc6d6f4e60e402932a3306783e062cc8683f364a7fc94bf238cacbab9) supposed to be in this PR? If so probably worth adding a comment into this PR mentioning it since it looks like a change in functionality.

Yes, I think this has already been merged, so actually went in with the changes for JDK8/11 as they were having the issues. , as per #961

steelhead31:
Added a commit to allow a new push of JDK/JRE17 due to the issues with GPG signing yesterday, and a need to republish corrected binaries. ( raised in issue #962 ), but not present in every RPM based distribution.

sxa (approved):

PR #965, opened 2024-07-26
JULY CPU: Re-introduce Linux Packages For s390x For JDK22

steelhead31:
s390x has been cleared for release, having been removed in the previous CPU.


sxa (approved):

sxa:
Noting that while s390x was not in the original release, there was a `.1` release in the last cycle that included s390x with an additional patch backported which is why it had been removed.

andrew-m-leonard (approved):
lgtm

PR #1004, opened 2024-09-16
implement jinja templating to handle linux installer updates

gdams:
experiment with using jinja to template linux spec files

I've started with alpine as it's the simplest to template

### DIFF
--- a/linux/jre/alpine/src/main/packaging/temurin/21/APKBUILD
+++ b/linux/jre/alpine/src/main/packaging/temurin/21/APKBUILD
@@ -1,11 +1,7 @@
 # Maintainer: Eclipse Adoptium Package Maintainers <temurin-dev@eclipse.org>
 pkgname=temurin-21
 pkgver=21.0.4_p7
-# replace _p1 with _1
-_pkgver=${pkgver/_p/_}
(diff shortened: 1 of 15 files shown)


linux/jre/alpine/src/main/packaging/temurin/21/APKBUILD
```diff
@@ -60,6 +61,7 @@ _jre() {
 	mv "$_fromroot/conf"    "$_toroot"
 	mv "$_fromroot/legal"   "$_toroot"
 	mv "$_fromroot/lib"     "$_toroot"
+	mv "$_fromroot/man"     "$_toroot"
```
gdams, line 64: @steelhead31 do we know if these were intentionally removed from more recent JRE's?

gdams: NVM, I've worked out why we skipped it now

steelhead31:
@gdams, oddly,  I've been looking at something similar today, just some bash scripting at the moment...

steelhead31:
@gdams see : https://github.com/steelhead31/installer/tree/sfrautomate/linux/sfr.dev

PR #1017, opened 2024-10-01
Add Java Version to versionless provides for Temurin debian installers

### LINKED ISSUE #1001: `java-runtime` version information missing in debian packaging format
debian packaging formats settled on providing a specific version of the `java-runtime` to make it easier for other applications to depend on the specific version of java without needing to depend on a particular java distributor.

temurin java packages include a provides for `java-runtime` however they don't specify the version (eg: `java-runtime (= 8)` such as seen on openjdk builds by debian/ubuntu)). This packaging bug makes it so versioned dependencies (eg: `Depends: java-runtime (= 8)` or  `Depends: java-runtime (>= 8)`) will never resolve any temurin packages as compatible. This effectively makes the inclusion of `Provides: java-runtime` useless without the version.

See https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides for further explanation

steelhead31:
Fixes #1001 

Include Java version details for Debian versionless packages.


tellison (approved):

sxa:
FYI @judovana in case you have any comments on this :-)

gdams (approved):

karianna (approved):

PR #1084, opened 2024-12-18
Investigate JDK8/JRE8 Alpine Action Failure

steelhead31:
The Github Action for JDK/JRE8 for alpine is failing

This is due to Alpine:latest container moving to 3.21, which causes issues for JDK/JRE8

This PR pins the dockerfile to 3.19.


adamfarley (approved):
LGTM

sxa (approved):
LGTM - can you add in a comment (just in this PR, not the code) mentioning what the failure message is so we can identify it if we see it in the future?
I'm assuming it's not a concern for an end user installing on Alpine 3.21?

steelhead31:
The error in the GHA shows as this.. which appears to be because of a dependency no longer present in Alpine 3.21,  this dependency is only required at build time, and is related to the apk build command. This doesn't impact the installation of a JDK built on an older version, deployed on Alpine 3.21 ) tested in an 3.21 alpine container.

```
>>> ERROR: temurin-8-jdk*: create_apks failed
>>> ERROR: temurin-8-jdk: rootpkg failed
```

```
> Task :jdk:alpine:packageJdkAlpine FAILED
FAILURE: Build failed with an exception.

* Where:
Build file '/home/runner/work/installer/installer/linux/jdk/alpine/build.gradle' line: 104

* What went wrong:
Execution failed for task ':jdk:alpine:packageJdkAlpine'.
> Process 'command 'docker'' finished with non-zero exit value 1

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':jdk:alpine:packageJdkAlpine'.
```

PR #1075, opened 2024-12-16
Add Ubuntu 24.10 Oracular Oriole

Mailaender:
https://canonical.com/blog/canonical-releases-ubuntu-24-10-oracular-oriole

PR #1085, opened 2024-12-19
Bump package version for Debian republish

steelhead31:
Following the merge of https://github.com/adoptium/installer/pull/1075 

A new version of the debian packages is required to be published to make them available for Ubuntu 24.10


sxa (approved):
Bit of a shame that we need to bump this to publish an extra package. Will this republish the versions for all Debian/Ubuntus at this new `-1` level or only the new one?

steelhead31:
I've added a note to the installer epic, it'd be handy to have a new function to just update the metadata for existing debian packages, rather than having to push a new version, likewise with the rhel/suse packages, we could do something similar when we move to the symlink method of uploading, rather than the multiple copies.

karianna (approved):

Issue #1005, opened 2024-09-17
EPIC: Automate & Improve Production Of Linux Package Installers


steelhead31:
There are currently a number of issues when producing the Linux APKs, Debs & RPMs for the JDKs / JREs at the end of a release cycle:

Notably:

*   Currently take ~1 week at the end of a release
*   Quite painful for CPU when doing 5 major version releases
*   Lots of manual updates required to numerous files, which can be prone to error.
*   Lots of work to add new versions and O/S distributions
*   All architectures for a version have to be completed, before installer packages can be produced
*   Removing a single architecture (when required) is even more time consuming and error prone.
*   Currently doing platform specific installers is quite complicated

Proposal:

Ideally, the linux packages should be produced on a per version/architecture basis, and this should be done at the same time that the linux distribution tar balls are published to Github.

The new process, might be as follows:

When the release binaries are published to github via the jenkins refactor release tool :  

*   Run a subsequent process to create the package installers, for each dist/version/architecture combination.
*   Once the linux installer packages are produced, push them to artifactory

This will require the following tasks:

*   Create a generic script that can build an individual arch, version, distro combination
*   Dynamically generate the APK build, debian control and RPM spec files.
*   For the job that generates files, how do we keep track of those files (e.g. pushed into a repo?)
    *   Ideally tracked in a GitHub repository

Advantages of this approach :

*   The process doesn’t disrupt the existing steps to test the installers

Improvement Items That Can Be Added:

*   Improve handling of src rpms to be architecture specific, rather than a single huge file, with EVERY architecture contained.
*   For the past issue where pushing to JFrog has failed
    *   Need a follow-on verification test for whether an artifact was pushed successfully on JFrog
    *   Download the artifact from jfrog and perform a simple test.
*   Also looking at using symlinks to reduce the number of uploads to JFrog, though the technical work is ongoing
*   Add metrics for how long it takes artifacts to be pushed to JFrog, but when we do improve this, we will also look back at previous releases to see how big the improvement is
    *   Would also be nice to get metrics for the container images
*   AIX installers don’t exist at this time

List Of Tasks:

*   [x] #1019
*   [x] #1037
*   [x] Add JDK8 version handling support ( Part of 1037 )
*   [x] Add JDK8 ARM32 version handling support ( Part of 1037 )
*   [x] Amend jenkinsfile to have stage specific nodes.
*   [x] #1049
*   [x] #1069
*   [x] #1045
*   [x] #1072
*   [x] #1080
[...]

PR #1103, opened 2025-01-22
Improve jenkins error logging in installer jenkinsfile & fix typo

steelhead31:
The handling of none linux packages, has shown a typo in printlin rather than println.

Also as the upstream job that triggers the build of the new linux packages cannot be selective, I need to handle the additional parameters to reduce entries in the jenkins warning logs. These parameters will not be used, just added to the pipeline to remove the warning about unhandled parameters.

Part of #1005


steelhead31:
Done in error

PR #1107, opened 2025-01-23
BugFix : Update automated installer production to validate architecture per distro

steelhead31:
A bug was spotted during the release cycle, whereby it was incorrectly trying to determine which alpine node to build on for a non alpine package build.

This bug fix amends the node identification to be based on the type of packages ( alpine / or non alpine ) being built.

Part of #1005

### DIFF
--- a/linux_new/Jenkinsfile
+++ b/linux_new/Jenkinsfile
@@ -323,11 +323,18 @@ pipeline {
                         error("The Artifacts contain multiple architectures: ${extractedArchs}")
                     }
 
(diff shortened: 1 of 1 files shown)


sxa (approved):
Looks like this has the fix for the "Skipped parameter" warnings from https://github.com/adoptium/infrastructure/issues/2774#issuecomment-2607257280 too :-)

andrew-m-leonard (approved):

gdams (approved):

PR #1119, opened 2025-01-29
Add files required to consistently build Alpine APKs for January Release

steelhead31:
Add the usual APKs so that the previous method of building the APK files works, also pin the alpine builds to docker latest.


smlambert (approved):

sxa (approved):
As long as the removal of the `/` in:

> `def fullURL = "${baseURL}/${binaryRepo}"`

is safe than I'm good with this - as far as I can tell there's nothing otherwise risky here

PR #1120, opened 2025-01-30
force pull base image before building docker image

gdams:
Also re-pin alpine image to 3.20. (Note that 3.21 currently breaks the JDK8 APKBUILD process)


steelhead31 (approved):
Looks good.

Issue #1218, opened 2025-07-23
Testing on Debian:Buster (past EOL) is causing jdk:debian:checkJdkDebian to fail. This blocks all future Debian builds and PRs


jmjaffe37:
We (MIcrosoft) ran into this issue toward the end of our July release: Debian buster now fails to run `apt-get update` and this causes build checks and PR merge checks to fail. Specifically, the error occurs here:

AptOperationsTest > packageSuccessfullyInstalled(String, String) > packaging.AptOperationsTest.packageSuccessfullyInstalled(String, String)[4] FAILED
    org.opentest4j.AssertionFailedError at AptOperationsTest.java:59

This is because: Debian:buster is now a year past its EOL and its default package servers seem to have just been shut off

jmjaffe37:
Possible solutions:
1. Remove Debian:buster from supported distros (this is what I vote for)
2. Remove Debian:buster just from distros used to test when invoking `checkJdkDebian `
3. Modify the Dockerfile to point `apt-get update` to legacy package servers (I would not prefer this option)

jmjaffe37:
Option 1 is proposed in this PR: https://github.com/adoptium/installer/pull/1220
Option 2 is proposed in this PR: https://github.com/adoptium/installer/pull/1219

karianna:
I think we need both PRs, have approved both.

jmjaffe37:
@karianna, to be clear: option 2 is a subset of option 1. The difference: Option 1 also removes `debian:buster` from build support and from the READMEs and documentation. If you want both, that means that we should only merge option 1.

jmjaffe37:
As per our discussion offline, we will only be going forward with the PR that is option 1 since it has all of the code changes that option 2 has plus more

jmjaffe37:
Solution option 1 has been merged and the new `main` branch has successfully created Debian installers again. Closing this issue now

PR #1220, opened 2025-07-23
Completely removed support for Debian:buster

jmjaffe37:
This distro is over a year EOL and its default package servers went offline recently. This causes builds and PRs to fail since `checkJdkDebian` needs to run `apt-get update` on each docker-image that it tests on.

This PR implements option 1 described in this issue: https://github.com/adoptium/installer/issues/1218

### DIFF
--- a/docs/Guide_To_The_Linux_Installers.md
+++ b/docs/Guide_To_The_Linux_Installers.md
@@ -508,8 +508,7 @@ def deb_versions = [
 		"trixie",   // Debian/13
 		"bookworm", // Debian/12
 		"bullseye", // Debian/11
-		"buster", // Debian/10
-
+		"buster",   // Debian/10
 		"oracular"  // Ubuntu/24.10 (STS)
 		"noble",    // Ubuntu/24.04 (LTS)
 		"jammy",    // Ubuntu/22.04 (LTS)


sxa:
_(Also noting that there is a valid counter-argument to my earlier comment the other way, which is that if they have to move to the `archive` repository for the main packages, they should perhaps also adjust the adoptium repo to do `s/buster/bookworm` or similar, which would also still work and I think may be more consistent with what we do for Ubuntu STS releases)_

karianna:
Added PMC Agenda to this.  I think it's worthy of a debate there on our _"Do we support EOL platforms, and if so, under what circumstances"_ :-)

sxa (changes requested):
I'm going to put a block on this for now as we were unable to have the discussion at this week's PMC call so it will be deferred until next week and I want to avoid this being merged in the meantime :-)

jmjaffe37:
@sxa, @karianna, are there any updates on this change? As a reminder, pipelines that build deb files are currently broken without this change since `checkJdkDebian` will always error until a change is made

linux/Jenkinsfile
```diff
@@ -532,7 +532,6 @@ def uploadDebArtifacts(String buildArch, String Version) {
             "trixie",   // Debian/13
             "bookworm", // Debian/12
             "bullseye", // Debian/11
-            "buster",   // Debian/10
```
steelhead31: I think this should be left in the jenkinsfile, as its just used to publish packages in the artifactory repo.

linux_new/Jenkinsfile
```diff
@@ -24,7 +24,6 @@ def deb_distros = [
 "trixie", // Debian/13
 "bookworm", // Debian/12
 "bullseye", // Debian/11
-"buster",   // Debian/10
```
steelhead31: Likewise for this.

steelhead31 (changes requested):

.github/workflows/cacert-publish.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       - name: Upload deb file to Artifactory
         if: steps.check-deb.outputs.file_exists == 'false'
         run: |
-          DISTRO_LIST="trixie,bookworm,buster,oracular,noble,jammy,focal,bionic"
```
steelhead31: Also this needs to remain, so the certs still get published.

karianna:
@jmjaffe37 some review comments for you.

linux/Jenkinsfile
jmjaffe37: Fixed :)

linux_new/Jenkinsfile
jmjaffe37: Fixed :)

.github/workflows/cacert-publish.yml
jmjaffe37: Fixed :)

docs/Guide_To_The_Linux_Installers.md
```diff
@@ -507,6 +508,7 @@ def deb_versions = [
 		"trixie",   // Debian/13
 		"bookworm", // Debian/12
 		"bullseye", // Debian/11
+    "buster",   // Debian/10
```
sxa: The indendation here looks inconsistent with the other lines

sxa (commented):
Idendation issue, but I'll defer to @steelhead31 for the fnal review.

docs/Guide_To_The_Linux_Installers.md
```diff
@@ -508,7 +508,7 @@ def deb_versions = [
 		"trixie",   // Debian/13
 		"bookworm", // Debian/12
 		"bullseye", // Debian/11
-		"buster",   // Debian/10
+    "buster",   // Debian/10
```
steelhead31: Small indentation suggestion, but it shouldnt impact anything
```suggestion
		"buster", // Debian/10

```

steelhead31 (approved):
Indentation as above is inconsistent, but otherwise looks good. Approving

docs/Guide_To_The_Linux_Installers.md
jmjaffe37: @sxa, I noticed this in the GH UI as well, but it is consistent in my branch when I look at it locally through VS Code so I am unsure what it happening here

PR #1305, opened 2025-12-16
Use local WixToolset.Heat if provided

Luigi96:
Use local WixToolset.Heat if provided

### DIFF
--- a/wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
+++ b/wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
@@ -124,21 +124,26 @@ Get-ChildItem -Path .\ -Filter *.zip -File -Name | ForEach-Object {
 
 # Install wixtoolset.heat version $wix_version
 $wix_path_provided = -not [string]::IsNullOrEmpty($wix_path)
+# Always stage wix_extension alongside the script's working directory.
+$wix_extension_dir = Join-Path -Path (Get-Location).Path -ChildPath "wix_extension"
+if (Test-Path -Path $wix_extension_dir) {
+  Remove-Item -Path $wix_extension_dir -Recurse -Force
+}
+New-Item -ItemType Directory -Path $wix_extension_dir -Force | Out-Null
+
 if ($wix_path_provided) {
   if (-not (Test-Path -Path $wix_path)) {
     throw "Provided WixToolset.Heat path '$wix_path' does not exist."
   }
   Write-Host "Using provided WixToolset.Heat path: $wix_path"
-  mkdir wix_extension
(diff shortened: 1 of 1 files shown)


wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
```diff
@@ -118,12 +123,23 @@ Get-ChildItem -Path .\ -Filter *.zip -File -Name | ForEach-Object {
...
-$outFile = '.\wix_extension\wixtoolset.heat.' + $wix_version + '.zip'
-Invoke-WebRequest -Uri $sourceURI -OutFile $outFile
-Expand-Archive -Path "$outFile" -DestinationPath ./wix_extension/
+$wix_path_provided = -not [string]::IsNullOrEmpty($wix_path)
+if ($wix_path_provided) {
+  if (-not (Test-Path -Path $wix_path)) {
+    throw "Provided WixToolset.Heat path '$wix_path' does not exist."
+  }
+  Write-Host "Using provided WixToolset.Heat path: $wix_path"
+  mkdir wix_extension
```
joe-braley: Nit: `mkdir` is an alias in Powershell. This wraps a different command which does not behave exactly like the bash equivalent. Believe this is an alias for the cmd `New-Item -ItemType Directory`. 

Also, might be wise to determine if the directory exists and/or if a `parent` directory needs to be created as well (e.g., `mkdir -p ...`)

wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
```diff
@@ -20,6 +20,9 @@
 .PARAMETER jvm
     The JVM to be used. If not provided, the script will attempt to extract the JVM from the filename.
 
+.PARAMETER wix_path
+    The path to an existing WixToolset.Heat installation. If provided, the script will use this path instead of downloading WixToolset.Heat.
```
joe-braley, line 24: Q: Is WixToolSet.Heat the only package which requires downloading? Or would this, potentially, include other packages than the one in question?

wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
joe-braley: Does looks like `wix_extension` is meant to be created in same directory, reading the following line, but might be wise to verify. Not a blocking suggestion.

wix/SourceDir/CreateSourceFolder.AdoptOpenJDK.ps1
Luigi96: This script is just to get WixToolSet.Heat, we don't download any other packages in this script.

karianna (approved):
with Joe's comments fixed this is approved

PR #1337, opened 2026-02-02
Add Housekeeping step to artifact publish job.

### LINKED ISSUE #1338: Debian Packages Uploaded To Wrong File Path
As discovered here...  https://adoptium.slack.com/archives/C09NLQQAV/p1769863485587969

Occasionally, debian packages have been uploaded into the wrong path ( this shouldnt have caused issues, as they were also uploaded to the correct path, and the metadata had been updated correctly.

The root cause of this is due to the re-use of workspace directories on dynamically provisioned agents.

steelhead31:
Fixes #1338 

Add a housekeeping step, to remove retrieved artifacts, once published. Reruns of this job, would pull the artifacts down again.

This should avoid the issue on dynamic agents, whereby re-use can cause artifacts to be inadvertently uploaded to the wrong path.

Also correct an errant "," in the debian check and upload process.

Tested with the corrections : https://ci.adoptium.net/job/build-scripts/job/release/job/sfr-build-linux-package-modular/1159/

### DIFF
--- a/linux_new/Jenkinsfile
+++ b/linux_new/Jenkinsfile
@@ -977,11 +977,11 @@ stage('Publish Packages') {
     }
 // Publish Packages Stage - End
 // Housekeeping Step
-post {
-  always {
-    cleanWs(deleteDirs: true, disableDeferredWipeout: true)
-  }
-}
+    post {
+      always {
+        cleanWs(deleteDirs: true, disableDeferredWipeout: true)
+      }
+    }
 // End Housekeeping
   }
 // Stage Definition - End


linux_new/Jenkinsfile
```diff
@@ -976,6 +976,13 @@ stage('Publish Packages') {
 // End OF Script
     }
 // Publish Packages Stage - End
+// Housekeeping Step
+post {
+  always {
+    cleanWs(deleteDirs: true, disableDeferredWipeout: true)
+  }
+}
```
adamfarley:
```suggestion
    post {
      always {
        cleanWs(deleteDirs: true, disableDeferredWipeout: true)
      }
    }
```
Indentation suggestion. What do you think @steelhead31 ?

PR #1383, opened 2026-04-08
Forcibly remove headful library files in headless JRE 26

Haroon-Khel:
These library files wont be present in headless installer builds, so the `-f` should cover both cases (when the files are and aren't present)

### DIFF
--- a/linux_new/jre/rhel/src/main/packaging/temurin/26/temurin-26-jre.template.j2
+++ b/linux_new/jre/rhel/src/main/packaging/temurin/26/temurin-26-jre.template.j2
@@ -155,7 +155,7 @@ echo "This is the headful version, nothing will be removed"
 %else
 echo "This is the headless version, headful libraries will be removed"
 pushd "%{buildroot}%{prefix}/lib"
- rm -v libsplashscreen.so libawt_xawt.so libjawt.so
+ rm -vf libsplashscreen.so libawt_xawt.so libjawt.so
 popd
 %endif
 


andrew-m-leonard (approved):
lgtm

Suggest adding PR description though

jerboaa (approved):

judovana:
> These library files wont be present in headless installer builds, so the `-f` should cover both cases (when the files are and aren't present)


Hello!  I have to be missing something. Both headless rpms and headfull rpms, are repacking the same blob.
If the files are in the headfull blob, they must be in headless-processing blob, and so must be removed. Maybe they got just moved elsewhere in jdk26? Or renamed?

judovana:
```
$ tar --list -f OpenJDK26U-jre_x64_linux_hotspot_26_35.tar.gz | grep -e libsplashscreen.so -e libawt_xawt.so -e libjawt.so
jdk-26+35-jre/lib/libsplashscreen.so
jdk-26+35-jre/lib/libjawt.so
jdk-26+35-jre/lib/libawt_xawt.so
$  tar --list -f OpenJDK26U-jdk_x64_linux_hotspot_26_35.tar.gz | grep -e libsplashscreen.so -e libawt_xawt.so -e libjawt.so
jdk-26+35/lib/libsplashscreen.so
jdk-26+35/lib/libjawt.so
jdk-26+35/lib/libawt_xawt.so
```

Quick check looks sane. The files must be removed, thus the `-f` was intentionally  missing. 
Maybe the path of `pushd "%{buildroot}%{prefix}/lib"` got changed?  Do you have some link to some failure?

jerboaa:
> > These library files wont be present in headless installer builds, so the `-f` should cover both cases (when the files are and aren't present)
> 
> Hello! I have to be missing something.

I think you do. The builds for riscv and some others - alpine? - are a headless JDK build (in contrast to the x64 Linux build). So they don't include headful bits in the result.

judovana:
Oh I see.  TY, "secondary arches", sure.  The missing `-f`  was some level of protection, maybe  maybe add `%if arch` condition, would be better?  Then different question, why jdk26 only? In such case, all jdk versions are affected.

sxa:
> Oh I see. TY, "secondary arches", sure. The missing `-f` was some level of protection, maybe maybe add `%if arch` condition, would be better? Then different question, why jdk26 only? In such case, all jdk versions are affected.

I'm speculating but it's likely just because 26 was the only version in the current release cycle and it was caused by something implemented in our packaging since January. I expect it would show up on other releases too.

Alpine and Linux/riscv64 are the only platforms we currently have which are headless-only (Although I've built and used a headfull JDK on RISC-V too in the past)

judovana:
> > Oh I see. TY, "secondary arches", sure. The missing `-f` was some level of protection, maybe maybe add `%if arch` condition, would be better? Then different question, why jdk26 only? In such case, all jdk versions are affected.
> 
> I'm speculating but it's likely just because 26 was the only version in the current release cycle and it was caused by something implemented in our packaging since January. I expect it would show up on other releases too.
> 

TY!  So once it bites, it will get fixed:)

> Alpine and Linux/riscv64 are the only platforms we currently have which are headless-only (Although I've built and used a headfull JDK on RISC-V too in the past)

As alpine is not rpm based, the removal should be executed as:
pushd "%{buildroot}%{prefix}/lib"
```
%ifnarch riscv64
     rm -v libsplashscreen.so libawt_xawt.so libjawt.so
%endif
  echo "Riscv64 should be headless by principle"
  if ls . | grep   -e libsplashscreen.so -e libawt_xawt.so -e libjawt.so; then
    echo "headful libraries present!" && exit 1
  fi
%endif
popd
```
Worthy of update? Which JDKs builts on riscv64?

sxa:
> Worthy of update?

I personally have no particular feelings one way or another.

> Which JDKs builts on riscv64?

17 and later

judovana:
Just minor note, jdk 25,21 and 17 have `rm -vf` since one of initials commits.   jdk26 missed it  by accident: https://github.com/adoptium/installer/commit/f2d9858df37c2695ff1207054ac3622ab01b3a42

PR #1458, opened 2026-08-28
Modernized README.md for wix, removed localization in URLs

jmjaffe37:
Removed both 'fr-fr' and 'en-us' localization strings from URLs, removed references to windows 10, and modernized the URLs (docs.microsoft.com -> learn.microsoft.com)

### DIFF
--- a/wix/README.md
+++ b/wix/README.md
@@ -140,7 +140,7 @@ Note:
 - FeatureOracleJavaSoft can't and must not be used per user install as it only write to HKLM. ([see details](https://docs.oracle.com/javase/9/install/installation-jdk-and-jre-microsoft-windows-platforms.htm#JSJIG-GUID-47C269A3-5220-412F-9E31-4B8C37A82BFB)
 - Machine PATH is always loaded before User PATH (FeatureEnvironment) ( If another java is installed per machine it will be the default one when using the PATH )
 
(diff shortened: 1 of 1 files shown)


d3r3kk (approved):
Ship it

joe-braley (approved):

karianna (approved):

karianna:
Weird that 25 and 8 pass but not the other 3.  @jmjaffe37 is that expected?

jmjaffe37:
@karianna, I did not look into the issue, but I saw that they failed within a minute, so I assume that it was a transient issue on the side of the runners/agents. I would imagine that running these again will result in success, but I do not seem to have permissions to rerun those tests.

Alos, since this change only effects the readme, it should be fine to merge either way

karianna:
@jmjaffe37 It continued to fail on repeated runs: Copilot thinks the fix should be as per https://github.com/adoptium/installer/pull/1460/changes - I _think_ that's correct given we've just had the version change to add the extra number but would appreciate your expert eye.

jmjaffe37:
@karianna, thanks for looking deeper! I just finished implementing a fix to the root problem (the github actions yml was not properly detecting the patch version). Please find my PR here (all automations passed) https://github.com/adoptium/installer/pull/1462

PR #1462, opened 2026-08-31
Fixed issue where automation did not properly pickup patch version

jmjaffe37:
Previously, the automation was always setting the input patch number to 0 because most JDKs did not have a patch number, so the API json response in the powershell code would return a null value for the patch number.

Now, if the response is null for the patch number, then it will be set to 0. Else, it will be set to the actual patch number.

I also added another way to find the correct `REPRO_DIR`, and documented this in the README.md


jmjaffe37:
@karianna, this is the PR that fixes the windows automation issue :)

karianna (approved):

d3r3kk (approved):
🚢 it

